PR #6, opened 2015-02-02
btrfs-convert: optout build

quinte17:
btrfs-convert has dependencies which won't be met in every circumstance.
This patch introduces a new make switch for it:
	DISABLE_BTRFSCONVERT=1 make

This patch is just to get the idea. There are still some targets missing.
I needed this option because I didn't need to convert and didn't have the dependencies at hand.

Suggestions/Merging welcome.


kdave:
The build system is currently getting converted to autoconf so I can't take the patch directly but I'll implement the same functionality on top of it.

NOTE: Please send patches to the btrfs mailinglist instead.

PR #7, opened 2015-07-02
btrfsck repair for I_ERR_FILE_WRONG_NBYTES and test case

adam900710:
Add repair function for I_ERR_FILE_WRONG_NBYTES and a test case for it.


kdave:
Applied directly, thanks.

PR #8, opened 2015-10-06
print root item's last_snapshot value.

Lakshmipathi:
Include last_snapshot value in print_root(). With btrfs-debug-tree helps to identify whether its a snapshot-ed subvolume or not.


kdave:
Thanks, the change is fine. Please send the patch via mailinglist, see https://btrfs.wiki.kernel.org/index.php/Btrfs_source_repositories#Note_to_GitHub_users for more.

Lakshmipathi:
thanks for the update and link, will send the patch via mailing-list.

Lakshmipathi:
Patch sent to the mailing-list. Will close this PR on github.

PR #11, opened 2016-01-13
Silence

moviuro:
This PR should be the first in a series of changes aiming at making `btrfs(8)` more script-friendly.
In here we remove the most annoying and noisy outputs.


moviuro:
meh, wrong branch I assume

PR #12, opened 2016-01-13
Silence

moviuro:
This PR should be the first in a series of changes aiming at making `btrfs(8)` more script-friendly.
In here we remove the most annoying and noisy outputs.

### DIFF
--- a/Documentation/btrfs-subvolume.asciidoc
+++ b/Documentation/btrfs-subvolume.asciidoc
@@ -43,7 +43,7 @@ from normal directories.
 
 SUBCOMMAND
 -----------
-*create* [-i <qgroupid>] [<dest>]<name>::
+*create* [-v] [-i <qgroupid>] [<dest>]<name>::
 Create a subvolume <name> in <dest>.
 +
 If <dest> is not given, subvolume <name> will be created in the currently
@@ -54,6 +54,9 @@ directory.
 -i <qgroupid>::::
 Add the newly created subvolume to a qgroup. This option can be given multiple
 times.
++
+-v::::
+be a bit more verbose
 
 *delete* [options] <subvolume> [<subvolume>...]::
 Delete the subvolume(s) from the filesystem.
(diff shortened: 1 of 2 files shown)


kdave:
Good idea in general, however this is a user visible change and likely to break scripts. Ideally we should rework the verbose output in one go. I'll look at it after 4.4 release.

moviuro:
My take on the issue was: commands should be silent in general. Hence "changing" the default behavior to a more standard one.
Also, the use of `btrfs(8)` in script is pretty complex because of some inconsistencies in its outputs (that's for an other PR). We could try to study who uses the verbose output in their projects (e.g. searching on github for people parsing `btrfs sub create`'s output)

On a side note, I am working on https://github.com/moviuro/butter which hit a number of weird things that `btrs-progs(8)` do, hence my contributions.

Do you have any guidelines as to how I should continue to silence the commands?

kdave:
My idea of the end result, regardless of the transition to the new behaviour:
- a command is silent unless it would be confusing what it did, ie. a single action like 'subvol create' is silent, 'device add' would print that it does trim (because it delays the whole action)
- the return code is 0 if ok, other in case of errors
- all commands get --verbose, and print what it does, repeated options are more verbose
- all commands get --quiet, disable everything except errors, in the example above 'device add' will not say anything about the trimming

Transition period would be counted in years (1-2), which might sound silly, but deprecation is always counted in years. Warning to users about the unexpected command silence needs to be communicated.

Incrementally we can start adding the --verbose/--quiet options right now.

moviuro:
Okay, so let's add `-q`/`-v` everywhere we can so that `btrfs(8)` doesn't complain about them being unknown options.

I'd also suggest switching to getopts everywhere (remember the `--` inconsistency?)

kdave:
Yeah, getopts should be used everywhere. I did not want to do that change just before the release. Please don't forget to send the patches to the mailinglist once you have them ready.

moviuro:
@kdave on a related note: how to I subscribe/register?

kdave:
https://btrfs.wiki.kernel.org/index.php/Btrfs_mailing_list

PR #14, opened 2016-05-10
From: Nicholas D Steeves <nsteeves@gmail.com>

sten0:
Date: Mon, 09 May 2016 20:04:57 -0400

Trivial fix for typos in comments and strings.

### DIFF
--- a/btrfs-convert.c
+++ b/btrfs-convert.c
@@ -2656,7 +2656,7 @@ fail:
 		close(fd);
 	if (is_btrfs)
 		fprintf(stderr,
-			"WARNING: an error occured during chunk mapping fixup, filesystem mountable but not finalized\n");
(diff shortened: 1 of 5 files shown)


kdave:
Can you please fix more typos? There should be some overlap with the kernel-source typos as well so you can use the same wordlist.

sten0:
On 11 May 2016 at 07:47, kdave <notifications@github.com> wrote:

> Can you please fix more typos? There should be some overlap with the
> kernel-source typos as well so you can use the same wordlist.
>
Done.  I sent the patch to linux-btrfs, because I read on the wiki that
github pull requests weren't the normal way of doing things.  For future
reference, are they ok, or are mailing list patches prefered?

kdave:
Patches to mailinglist are preferred. Pull request can work as a preview though.

PR #19, opened 2016-12-28
Fix ASSERT() logic when backtrace is disabled

ebfe:
dd2c2a4c67d509d90142b9443aa7449ef4257b2b apparently flipped the logic
of ASSERT() for the BTRFS_DISABLE_BACKTRACE case which results in bogus
assertion failures if configured with --disable-backtrace.


ebfe:
Fixed in 4.9.1.

PR #28, opened 2017-03-12
btrfs-progs: add missing limits headers

baruchsiach:
Build under musl libc fails because of missing PATH_MAX and XATTR_NAME_MAX
macro declarations. Add the required headers.

Signed-off-by: Baruch Siach <baruch@tkos.co.il>


kdave:
Applied, thanks. I'd like to add build verification in musl-based environment, will open an issue.

PR #39, opened 2017-04-07
Fix typo

gabo77:


kdave:
Applied, thanks. I wrote some changelog, let me know if you have objectios. I want to keep the level of "formalities", but for such a trivial typo I won't ask you to rework the branch.

PR #46, opened 2017-06-03
Trump docs

lukaslueg:


kdave:
The media nonsense does not need to be stored in the git history, fixed in another way.

Issue #50, opened 2017-06-18
<dev> ?== <devid> ?== <device>


Kasijjuf:
In reading the file [btrfs-device.asciidoc](https://github.com/kdave/btrfs-progs/blob/devel/Documentation/btrfs-device.asciidoc) I noticed the parameter notation for the subcommands varies from command to command. For example:

_add_ accepts one or more **\<dev>**
_remove_ accepts one or more **\<dev>** or **\<devid>**
_ready_ accepts a single **\<device>**

What is the difference between **\<dev>**, **\<devid>**, and **\<device>**?

kdave:
dev is supposed to be a path to device  (/dev/sda1 etc), devid is the numeric identifier of a device within the filesystem

kdave:
We could unify the naming in all pages and document that in the "Command syntax" section to the main page. So I suggest to drop the short '<dev>' and replace with '<device>'.  Please let me know if you're going to send a patch/pull for that (otherwise I'll do that myself).

Kasijjuf:
> dev is supposed to be a path to device (/dev/sda1 etc), devid is the numeric identifier of a device within the filesystem

What about **\<device>** then? Any different than **\<dev>**?


> So I suggest to drop the short '' and replace with ''.

What? I think the markdown parser ate your sentence. Did you mean to say:

"
So I suggest to drop the short ''\<dev>" and replace with ''\<device>".
"

kdave:
Hehe, you're right about both. The idea is to replace shortened dev with device.

PR #52, opened 2017-06-22
btrfs-progs: docs: replace <dev> with <device>

Kasijjuf:
as discussed in #50

I believe this replaces all instances of \<dev>. Please confirm.


kdave:
Applied, thanks.

PR #65, opened 2017-09-12
Tracking: extent buffer caching

kdave:
Add extent_buffer caching support. This will speedup check.


kdave:
Right now crashes on fsck/003-shift-offsets

kdave:
Released in 4.13.2.

PR #66, opened 2017-09-12
Tracking: mkfs --rootdir rework (v3)

kdave:
The mkfs --rootdir uses own chunk allocation and writes to the forbidden areas. This patchset builds on top of existing  mkfs code. The behaviour regarding final image size may change, a fallback option for minimizing the result will be provided.


kdave:
Updated to v3

kdave:
Partially merged.

kdave:
Merged to devel.

PR #73, opened 2017-10-24
btrfs-progs: defrag: add a brief warning about ref-link breakage

firegurafiku:
```text
There is a warning in btrfs-filesystem(8) saying that running 'defrag'
in Linux will almost certainly break ref-links, with much data potentially
being physically duplicated.

However, many users tend to read man pages *after* trying to run things
on their own risk and may miss this important information. This commit
adds a brief copy of this warning into the command built-in help message
where it has good chances to be spotted before user is stuck with
a crowded filesystem.

(Signed-off-by)
```

If think it's a documentation-only change, so I decided to skip the mailing list routine. I'm new to Btrfs; in case I misunderstood something, feel free to reject the PR.
BTW, could someone please answer https://unix.stackexchange.com/questions/400225?

### DIFF
--- a/cmds-filesystem.c
+++ b/cmds-filesystem.c
@@ -893,6 +893,11 @@ static const char * const cmd_filesystem_defrag_usage[] = {
 	"-s start            defragment only from byte onward",
 	"-l len              defragment only up to len bytes",
 	"-t size             target extent size hint (default: 32M)",
+	"",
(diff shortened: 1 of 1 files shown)


kdave:
Pull requests are ok for documentation (this is mentioned in README, I could make it more visible though). The text added to the command help looks ok, short enough and points to the manual page. Merged to devel.

kdave:
Released.

PR #80, opened 2017-11-26
Update btrfs-subvolume.asciidoc

TopView:
To simplify, I suggest moving the 'writable/readonly' issue only to the -r line, instead of having it introduced in two places.


kdave:
Merged, thanks.

PR #83, opened 2017-12-15
Btrfs-progrs: make defrag -c idempotent operation

Nefelim4ag:
Currently -c assume that user want compress data,
but sometimes user want uncompress data.

defrag_ioctl understood comrpession type and BTRFS_DEFRAG_RANGE_COMPRESS.

Use BTRFS_DEFRAG_RANGE_COMPRESS as signal that caller want
change compression type, and set compression type to none.

So allow do:
-cnone or -cno

Signed-off-by: Timofey Titovets <nefelim4ag@gmail.com>


Nefelim4ag:
That need small in kernel updates to work (I testing patches),
the base idea is to use BTRFS_DEFRAG_RANGE_COMPRESS as flag "we want change compression type", instead of use compress_type for that.

Issue #92, opened 2018-02-13
Tracking issue: gcc + asan test coverage


kdave:
What tests pass with `make D=asan CC=gcc` build. The list is edited in-place, updates as comments.

- [x] test-cli
- [x] test-mkfs
   - [x] 009-special-files-for-rootdir - minor leak
- [x] test-misc
   - [x] 005-convert-progress-thread-crash - leaks
   - [x] 008-leaf-crossing-stripes - leaks
   - [x] 010-convert-delete-ext2-subvol - leaks
- [x] test-check
   - [x] 020-extent-ref-cases - heap overflow

adam900710:
020 overflow fixed:
https://patchwork.kernel.org/patch/10218105/

adam900710:
mkfs:  009-special-files-for-rootdir - minor leak fixed:
https://patchwork.kernel.org/patch/10218129/

adam900710:
misc: 008
https://patchwork.kernel.org/patch/10218147/

kdave:
Excellent, thanks. Patches applied. I'll do the pre-release asan checks on my machine, the travis CI would need more work to get that automated, but that's for another issue.

PR #95, opened 2018-02-14
Asan fixes

adam900710:
Asan fixes for all the problems described at https://github.com/kdave/btrfs-progs/issues/92.

Rebased to devel branch, so the first patch may be a little different than the one in maillist, which is based on v4.15.


kdave:
Merged, thanks.

Issue #96, opened 2018-02-14
"btrfs send -p" fails if source and parent subvolumes are on different mountpoints (memory corruption)


digint:
If source and parent subvolumes are on different mountpoints, `btrfs send -p <parent> <source>` fails with memory corruption.

In cmds-send.c, it is wrongly assumed that source and parent share the same mountpoint [1], leading to subvol_strip_mountpoint() returning a char* pointer which can exceed the length of full_path (no bounds checking) [2].

  [1] https://github.com/kdave/btrfs-progs/blob/v4.15/cmds-send.c#L657
  [2] https://github.com/kdave/btrfs-progs/blob/v4.15/utils.c#L2490

Steps to reproduce:

```
# mkdir /tmp/btrfs_unittest
# truncate -s 200M /tmp/btrfs_unittest/btrfs0
# losetup /dev/loop0 /tmp/btrfs_unittest/btrfs0
# mkfs.btrfs -L BTRFS_UNITTEST /dev/loop0

# mkdir /tmp/btrfs_unittest/mnt_root
# mount /dev/loop0 /tmp/btrfs_unittest/mnt_root

# mkdir /tmp/btrfs_unittest/mnt_snapshots
# btrfs sub create /tmp/btrfs_unittest/mnt_root/snapshots
# mount -o subvol=snapshots /dev/loop0 /tmp/btrfs_unittest/mnt_snapshots

# btrfs sub create /tmp/btrfs_unittest/mnt_root/test0
# btrfs sub snap -r /tmp/btrfs_unittest/mnt_root/test0 /tmp/btrfs_unittest/mnt_snapshots/test0_snap_mnt_snapshots
# btrfs sub snap -r /tmp/btrfs_unittest/mnt_root/test0 /tmp/btrfs_unittest/mnt_root/test0_snap_mnt_root

# btrfs send -p /tmp/btrfs_unittest/mnt_snapshots/test0_snap_mnt_snapshots /tmp/btrfs_unittest/mnt_root/test0_snap_mnt_root > /dev/null
ERROR: open hots/test0_snap_mnt_snapshots failed. No such file or directory
ERROR: could not resolve rootid for /tmp/btrfs_unittest/mnt_snapshots/test0_snap_mnt_snapshots
```

digint:
Added pull request: #98 

With the patch applied, the test above gives:
```
[...]

PR #98, opened 2018-02-17
btrfs-progs: prevent incorrect use of subvol_strip_mountpoint

digint:
Add additional bound checks to prevent memory corruption on incorrect
usage of subvol_strip_mountpoint. Assert sane return value by properly
comparing the mount point to the full_path before stripping it off.

Mitigates issue: "btrfs send -p" fails if source and parent subvolumes
are on different mountpoints (memory corruption):

    https://github.com/kdave/btrfs-progs/issues/96

Note that this does not properly fix this bug, but prevents a possible
security issue by unexpected usage of "btrfs send -p".

Signed-off-by: Axel Burri <axel@tty0.ch>

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -2484,6 +2484,12 @@ const char *subvol_strip_mountpoint(const char *mnt, const char *full_path)
 	if (!len)
 		return full_path;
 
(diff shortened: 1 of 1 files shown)


kdave:
Applied, thanks.

PR #106, opened 2018-03-12
Remaing check code split

adam900710:
The head is d0deb7b44cc6b787287fa8165aafd9587f92f418, with the new inline extent size fix applied as the first patch.


kdave:
> with the new inline extent size fix applied as the first patch
Please don't do that, it does not make sense to revert a patch that's still in devel and breaks tests. I've folded the two patches together. The rest looks ok.

kdave:
And the patch also breaks tests-misc/002-uuid-rewrite. I'll move that out of devel for now so I can merge the 3 patches.

kdave:
Aha, no there's the dependency so all patches have to be merged at once. Please send me something that passes the relevant tests.

adam900710:
Didn't expect misc would be related, sorry for that.
I just ran check and convert and nothing wrong found.

And as you can see, such code split is very dependent on the check code base.
Folding code into previous patch would make things pretty hard to merge.

Would you please provide one stable HEAD for me to work?
Current rebase is all done manually as the moved code can't be renewed automatically.

adam900710:
Problem located.
Since we don't modify BTRFS_MAX_INLINE_DATA_SIZE() macro in the latest fix, we need to modify all callers to do correct check.
Extra check in mkfs/rootdir.c and convert/source-reiserfs.c is missing, leading to such problem.

I'd like to push the fix first and halt the split for a while.

kdave:
Stable head for development is hard to achieve when there are so many things in flux. The fix and code split came at a bad time. Right now I want to do a bugfix release so the code split is going to be postponed.

adam900710:
I'm fine with that.

I could wait for a less active cycle for the remaining part.

2018年3月19日 21:53，"kdave" <notifications@github.com>写道：

Issue #118, opened 2018-03-28
misc-test 006 sometimes fails when UUID tree is created at mkfs time


t-msn:
With the patch fa754dddc204456083fb2c44c69345a978feb2b5 ("btrfs-progs: mkfs: precreate the uuid tree", currently in devel), misc-test 006 occasionally fails at misc-test 006 (kernel is 4.16.0-rc7):

```
$ sudo make test-misc TEST=006\*
    [LD]     fssum
    [TEST]   misc-tests.sh
    [TEST/misc]   006-image-on-missing-device
/usr/data/src/btrfs-progs/tests//common: line 177: 10819 Aborted
                 (core dumped) $INSTRUMENT "$@" >> "$RESULTS" 2>&1
mayfail: returned code 134 (SIGABRT), not ignored
test failed for case 006-image-on-missing-device
make: *** [Makefile:329: test-misc] Error 1
```

Full log of misc-tests-resut.txt when the test fails is below:
```
=== Entering /usr/data/src/btrfs-progs/tests//misc-tests/006-image-on-missing-device
############### losetup --find --show img1
/dev/loop0
############### losetup --find --show img2
/dev/loop1
############### /usr/data/src/btrfs-progs/mkfs.btrfs -f -d raid1 -m raid1 /dev/loop0 /dev/loop1
btrfs-progs v4.15.1
See http://btrfs.wiki.kernel.org for more information.

Performing full device TRIM /dev/loop0 (2.00GiB) ...
Performing full device TRIM /dev/loop1 (2.00GiB) ...
Label:              (null)
UUID:               220272d0-2454-429c-a198-83eaf954aceb
Node size:          16384
Sector size:        4096
Filesystem size:    4.00GiB
Block group profiles:
  Data:             RAID1           204.75MiB
  Metadata:         RAID1           204.75MiB
  System:           RAID1             8.00MiB
SSD detected:       no
Incompat features:  extref, skinny-metadata
Number of devices:  2
Devices:
   ID        SIZE  PATH
    1     2.00GiB  /dev/loop0
    2     2.00GiB  /dev/loop1

############### mount /dev/loop0 /usr/data/src/btrfs-progs/tests//mnt
############### dd if=/dev/zero of=/usr/data/src/btrfs-progs/tests//mnt/a bs=1M count=10
10+0 records in
10+0 records out
10485760 bytes (10 MB, 10 MiB) copied, 0.00643606 s, 1.6 GB/s
[...]

PR #119, opened 2018-03-30
Btrfs image fix + eb cleanup

adam900710:
Fix for https://github.com/kdave/btrfs-progs/issues/118

In short, the problem is caused by some old code (read_extent_data()
from ancient btrfs check code) and offset-by-one from btrfs-image.
Which makes btrfs-image can only read from the first stripe of RAID1.

And if device of the first stripe is missing, btrfs-image will fail.

Fix the problem and add test case for it.

Along with extent buffer cleanup (refactors), which looks pretty harmless.

test-fsck, test-misc and test-mkfs passed. (test-convert is a little time consuming, but it's not related)


kdave:
Merged to devel, thanks.

PR #124, opened 2018-04-25
traverse_directory: Reset error code on continue.

benkard:
In case add_inode_items returned -EEXIST, traverse_directory would
handle the condition and still continue under certain circumstances, but
it would not reset the error code, leading to spurious failure later.
This patch fixes that.


kdave:
Merged to devel, thanks.

PR #125, opened 2018-04-27
SLE11 build fixes

jeffmahoney:
This patch set allows btrfs-progs to build without further patching
on a stock SLE11 SP4 install.  The major issues were:
- new dependencies on e2fsprogs 1.42
- use of -std=gnu90, which gcc 4.3.4 doesn't support
- use of m4_chomp, introduced in autoconf 2.64.


jeffmahoney:
Since we still require autoconf 2.64, the last commit is silly. I'll resubmit without it.

Issue #128, opened 2018-05-05
btrfs inspect-internal dump-tree segmentation fault


tomato42:
when running
```
btrfs inspect-internal dump-tree -b 58470772736 /dev/stacja/backup-old
```
on a damaged btrfs, I'm getting a segmentation fault.

this is both with `v4.16` and f7fc27cb20924cc340a2a522655969253dd97ae9

<details>
    <summary>standalone run</summary><p>

```
[root@bursa22 btrfs-progs]# LD_LIBRARY_PATH=. ./btrfs inspect-internal dump-tree -b 58470772736 /dev/stacja/backup-old
btrfs-progs v4.16.1
ctree.c:1628: leaf_space_used: Warning: assertion `data_len < 0` failed, value 1
./btrfs(+0x135b3)[0x560206aaa5b3]
./btrfs(+0x13662)[0x560206aaa662]
./btrfs(+0x18ae0)[0x560206aafae0]
./btrfs(btrfs_leaf_free_space+0x4f)[0x560206aafb4d]
./btrfs(btrfs_print_leaf+0xdc)[0x560206acb967]
./btrfs(btrfs_print_tree+0x8d)[0x560206acc155]
./btrfs(cmd_inspect_dump_tree+0x4d5)[0x560206b45275]
./btrfs(handle_command_group+0xae)[0x560206aaa058]
./btrfs(cmd_inspect+0x38)[0x560206b0ad6f]
./btrfs(main+0x138)[0x560206aaa43d]
/usr/lib/libc.so.6(__libc_start_main+0xe7)[0x7f1ce8b529a7]
./btrfs(_start+0x2a)[0x560206aa9c3a]
leaf 58470772736 items 84 free space 501708 generation 92338 owner EXTENT_TREE
leaf 58470772736 flags 0x1(WRITTEN) backref revision 1
fs uuid 8baae558-87e2-45ff-93dc-a03c8c3e8050
chunk uuid ae419504-168b-4614-9888-aab3e8d2f187
        item 0 key (71857111040 EXTENT_ITEM 4096) itemoff 3935 itemsize 60
                refs 66050 gen 67973 flags TREE_BLOCK|FULL_BACKREF
                tree block key (220466 DIR_INDEX 389405254) level 169
                shared block backref parent 75269525520
                shared block backref parent 70586298626
        item 1 key (71857123328 EXTENT_ITEM 4096) itemoff 3884 itemsize 51
[...]

PR #132, opened 2018-05-07
Dump tree segv fix

adam900710:
For Issue: #128 
With test case.


kdave:
Merged to devel (some time ago actually), thanks.

PR #135, opened 2018-05-14
btrfs-progs: Detect compressed extent without csum

adam900710:
It's based on v4.16 stable branch.

James Harvey from mail list reports a strange kernel panic, whichs show
obviously kernel memory corruption, while after btrfs decompression
failure.

It turns out that, some compressed extent get corrupted on-disk, while
the inode has NODATASUM set, there is no csum to prevent corrupted
mirror being used.

Although the root cause should be buggy lzo implementation, it still
shows that btrfs is not following the behavior defined in btrfs(5):

     Note
     If nodatacow or nodatasum are enabled, compression is disabled.

So at least make btrfs check to detect such problem.

Kernel patch sent titled:
"[PATCH] btrfs: inode: Don't compress if NODATASUM or NODATACOW set"


jamespharvey20:
If these files are in snapshots, is there any way to fix them, without deleting all the old snapshots?  Any chance of a future `--repair` taking care of these?  "Fixing" probably means if nodatacow or nodatasum are enabled, changing the data on disk to be uncompressed; and if neither are enabled now, leaving on disk compressed but generating csums.  Or, if too complicated, just changing to uncompressed in all scenarios.

kdave:
Added to devel.

PR #138, opened 2018-05-15
btrfs-progs: fix regression preventing send -p with subvolumes mounted on "/"

digint:
Fix subvol_strip_mountpoint for mnt="/" (len=1). In this case, skip
check for trailing slash on full_path (leading slash on full_path is
already asserted by strncmp).

Fixes: c5dc299aff6b ("btrfs-progs: prevent incorrect use of subvol_strip_mountpoint")
Signed-off-by: Axel Burri <axel@tty0.ch>


kdave:
Merged to devel, thanks.

PR #139, opened 2018-05-16
qgroups usability v3

jeffmahoney:
This is v3 of the qgroups usability patchset.

Changes since v2:
- Updated Reviewed-by tags were provided.
- Fixed the typoed commands that I commented on in the previous posting.
- Dropped the btrfs_cleanup_root_info since it's unnecessary with the
  switch to libbtrfsutil.
- Updated the qgroups pathname patch to use libbtrfsutil
- Added a mini-filter to skip dead qgroups in 'qgroup show' output
  unless -v is specified.

The most notable change is that this posting doesn't include the JSON
format patches since there was still some open discussion there.  The
plumbing for passing global options around to commands is there, so we
really just need to decide on how we want to handle alternative formats.
For some commands, the formatting library that coreutils uses will
probably work but for qgroups show to represent nested groups, it's
unsuitable.


kdave:
The patchset is doing several things at once, I've merged some of them over the time (command line parsing, fixes), the output --format as part of the global option rework. I'd rather track several PRs/issues so it's clear what's merged or needs to be fixed, this PR has become unwieldy.

kdave:
Remaining parts of the pull request have been rebased and merged:
- added -W option to rescan
- print path in qgroup show

PR #146, opened 2018-09-05
btrfs-progs: docs: update remaining non-zero dev stats -s option to -c

phillxnet:
Update the remaining erroneous entry of -s in the 'exit status' section to -c in accordance with the changes made in:

"btrfs-progs: dev stats: update option name for checking non-zero status"

Signed-off-by: Philip Guyton <philip@yewtreeapps.com>


kdave:
Thanks, I should have applied that long time ago sorry, now in devel.

PR #149, opened 2018-10-01
rename create_subvolume and create_snapshot arg 'async' to 'no_wait'

theridane:
Python 3.7 introduces a bunch of new keywords including `async` which breaks keyword arguments of two methods in `btrfsutil` (e.g. causing test_subvolume to fail with a syntax error).

This renames the arguments to `no_wait`.


kdave:
Was fixed by 5776a70b3024d03eeb4 .

PR #151, opened 2018-10-15
set up max_errors count

Alexander198961:
Try to set up max retry count to fix https://bugzilla.kernel.org/show_bug.cgi?id=200085


kdave:
Thanks, fix added to devel, with enhanced changelog.

PR #154, opened 2018-11-26
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

Note: I understand that the commit message style doesn't quite match. I can squish/rewrite as requested.

There are a couple of commits where I'm guessing about the corrections...

### DIFF
--- a/check/main.c
+++ b/check/main.c
@@ -9146,7 +9146,7 @@ static int build_roots_info_cache(struct btrfs_fs_info *info)
 		 * It's a valid extent/metadata item that has no inline ref,
 		 * but SHARED_BLOCK_REF or other shared references.
 		 * So we need to do extra check to avoid reading beyond leaf
-		 * boudnary.
+		 * boundary.
 		 */
 		if ((unsigned long)iref >= item_end)
 			goto next;
--- a/check/main.c
+++ b/check/main.c
@@ -4052,7 +4052,7 @@ static int try_to_fix_bad_block(struct btrfs_root *root,
 	btrfs_init_path(&path);
 	ULIST_ITER_INIT(&iter);
 	/*
-	 * If we found no roots referrening to this tree block, there is no
+	 * If we found no roots referencing to this tree block, there is no
 	 * chance to fix. So our default ret is -EIO.
 	 */
 	ret = -EIO;
--- a/check/main.c
+++ b/check/main.c
@@ -8654,7 +8654,7 @@ static int reinit_extent_tree(struct btrfs_trans_handle *trans,
 	 * first we need to walk all of the trees except the extent tree and pin
 	 * down/exclude the bytes that are in use so we don't overwrite any
 	 * existing metadata.
-	 * If pinnned, unpin will be done in the end of transaction.
+	 * If pinned, unpin will be done in the end of transaction.
 	 * If excluded, cleanup will be done in check_chunks_and_extents_lowmem.
 	 */
 again:
--- a/check/main.c
+++ b/check/main.c
@@ -7851,7 +7851,7 @@ static int check_chunk_refs(struct chunk_record *chunk_rec,
 			    dev_extent_rec->length != length) {
 				if (!silent)
 					fprintf(stderr,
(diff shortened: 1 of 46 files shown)


check/main.c
```diff
@@ -7851,7 +7851,7 @@ static int check_chunk_refs(struct chunk_record *chunk_rec,
 			    dev_extent_rec->length != length) {
 				if (!silent)
 					fprintf(stderr,
-						"Chunk[%llu, %u, %llu] stripe[%llu, %llu] dismatch dev extent[%llu, %llu, %llu]\n",
+						"Chunk[%llu, %u, %llu] stripe[%llu, %llu] mismatch dev extent[%llu, %llu, %llu]\n",
```
jsoref: I'm not sure about this

check/mode-lowmem.c
```diff
@@ -2636,7 +2636,7 @@ static int process_one_leaf(struct btrfs_root *root, struct btrfs_path *path,
 	if (err & LAST_ITEM)
 		goto out;
 
-	/* still have inode items in thie leaf */
+	/* still have inode items in the leaf */
```
jsoref: Possibly "this"?

convert/common.c
```diff
@@ -76,7 +76,7 @@ static inline int write_temp_super(int fd, struct btrfs_super_block *sb,
 }
 
 /*
- * Setup temporary superblock at cfg->super_bynter
+ * Setup temporary superblock at cfg->super_bytenr
```
jsoref: Pretty sure this is right...

send-stream.c
```diff
@@ -35,7 +35,7 @@ struct btrfs_send_stream {
 
 	/*
 	 * end of last successful read, equivalent to start of current
-	 * malformated part of block
+	 * malformed part of block
```
jsoref: this might be objectionable...

show-blocks
```diff
@@ -33,15 +33,15 @@ rcParams['backend'] = 'Agg'
 rcParams['interactive'] = 'False'
 from pylab import *
 
-class AnnoteFinder:
+class AnnotationFinder:
```
jsoref: It's possible that `annote` is intentional -- I can easily drop any individual commits

tests/fsck-tests/025-file-extents/test.sh
```diff
@@ -15,7 +15,7 @@ prepare_test_dev 128M
 
 # Do some write into a large prealloc range
 # Lowmem mode can report missing csum due to wrong csum range
-test_paritical_write_into_prealloc()
+test_partial_write_into_prealloc()
```
jsoref: Not sure about this

check/main.c
kdave: It's 'mismatch'

check/mode-lowmem.c
kdave: Yes, 'this'

convert/common.c
kdave: Indeed, refers to the member 'super_bytenr' in struct btrfs_fs_info

send-stream.c
kdave: I don't remember exactly, as the stream is a formatted stream of bytes, the 'malformatted' might make sense too, but 'malformed' is not cofusing in this context

show-blocks
kdave: The 'show-blocks' script seems to be stale, I don't remember anyone using it, so fixing there is not necessary as we might delete it anyway.

tests/fsck-tests/025-file-extents/test.sh
kdave: 'partial' looks correct to me

kdave:
Thanks for the fixes, no matter how much we try not to introduce them, they keep coming. The changes would be better grouped logically, eg. documentation, CHANGES, and code. I'll do that as the separate commits make it easy to cherry-pick and squash relevant changes, so if you have further updates, please keep it as is and I'll do the final merge.

jsoref:
@kdave : Thanks for offering to rearrange this PR. At present, I don't have any other changes -- my script ran through the entire corpus.

kdave:
Changes grouped into commits, now pushed to devel branch. Closing.

PR #158, opened 2018-12-16
Mention how to find the subvolume creation time

jondo:
As a btrfs newbie, having the description of _show_ list the shown fields (at least the creation time) would just have been very useful to me.


kdave:
Thanks for the suggestion, I did a few more updates in b4a2708d7160bc4a123bb4b66290e06faaa15138. There's an example output in the manual page that contains the creation time, I think this should be sufficient to make the creation time easy to find.

jondo:
Très cool, thank you!

PR #172, opened 2019-05-19
Makefile: apply LDFLAGS to libbtrfsutil.so

trofi:
libbtrfs.so already has user's LDFLAGS applied.
The change also applies those to libbtrfsutil.so.

Reported-by: Michał Górny
Bug: https://bugs.gentoo.org/686284


kdave:
Added to devel, thanks. I've added an extra variable for the libbtrfsutil, but the user-specified LDFLAGS will be passed the same way.

PR #176, opened 2019-05-27
btrfs-progs: Pass CFLAGS and LDFLAGS to Python

JPEWdev:
Adds Make variables EXTRA_PYTHON_CFLAGS and EXTRA_PYTHON_LDFLAGS which
can be used to pass CFLAGS and LDFLAGS respectively when building the
Python library.

This is required to support reproducible builds, as there are often
compiler and linker flags that must be passed in order to generate
reproducible output (e.g. -fdebug-prefix-map)

Signed-off-by: Joshua Watt <JPEWhacker@gmail.com>


kdave:
Added to devel, thanks.

PR #177, opened 2019-06-04
btrfs-progs: scrub: show the scrubbing rate and estimated time to finish

gkowal:
The estimation is based on the allocated bytes, so it might be
overestimated.

Example output for running scrub:

scrub status for xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
        scrub started at Fri May 31 15:56:57 2019, running for 0:04:31
        total 62.55GiB scrubbed at rate 236.37MiB/s, time left: 0:12:31
        no errors found

Signed-off-by: Grzegorz Kowal <grzegorz@amuncode.org>

### DIFF
--- a/cmds-scrub.c
+++ b/cmds-scrub.c
@@ -134,10 +134,16 @@ static void print_scrub_full(struct btrfs_scrub_progress *sp)
 		printf(" %s=%llu", desc, test);	\
 } while (0)
 
-static void print_scrub_summary(struct btrfs_scrub_progress *p)
+static void print_scrub_summary(struct btrfs_scrub_progress *p, struct scrub_stats *s, u64 bytes_total)
 {
 	u64 err_cnt;
 	u64 err_cnt2;
+	u64 bytes_scrubbed = p->data_bytes_scrubbed + p->tree_bytes_scrubbed;
+	u64 bytes_per_sec  = (s->duration > 0) ? (bytes_scrubbed / s->duration) : 0;
+	u64 seconds_left   = (bytes_per_sec > 0) ? (bytes_total - bytes_scrubbed) / bytes_per_sec : 0;
+	unsigned h_left = seconds_left / 3600;
+	unsigned m_left = (seconds_left / 60) % 60;
+	unsigned s_left = seconds_left % 60;
 
 	err_cnt = p->read_errors +
 			p->csum_errors +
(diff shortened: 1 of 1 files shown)


kdave:
I think the numbers match, scrub needs to go through all physical copies and this is sotred in the total and used _bytes in dev info.

Naturally the IO performance can vary so the average speed can be off if it's based on just one measurement, but as it is now I think it's ok.

kdave:
So the device info value 'total_used' cannot be used because it's the chunk size and not the amount of data used in the chunks. Ie. the same what's printed in 'fi df' as total, while we're interested in 'used'. This can be obtained by get_df.

I've reworked the output too so now it looks a bit more readable and adds some fancy stuff:

```
UUID:             bf8720e0-606b-4065-8320-b48df2e8e669
Scrub started:    Fri Jun 14 19:49:47 2019
Status:           running
Duration:         0:14:11
Time left:        0:04:04
ETA:              Fri Jun 14 19:53:51 2019
Total to scrub:   182.55GiB
Bytes scrubbed:   141.80GiB
Rate:             170.63MiB/s
Error summary:    csum=7
  Corrected:      0
  Uncorrectable:  7
  Unverified:     0
```

For the reference, this is 'fi df':

```
Data, single: total=261.00GiB, used=179.91GiB
System, single: total=32.00MiB, used=48.00KiB
Metadata, single: total=5.00GiB, used=2.64GiB
GlobalReserve, single: total=375.23MiB, used=0.00B
```

I've run it several times, the average time is 18 minutes and the estimate is not bad, usually off by one or two minutes.

gkowal:
I thought that even the overestimated time left using the size of allocated chunks is better than nothing, and was planning to find out later how to get the used space.

Anyway, on an active filesystem, the info about the used space does not guarantee more precise estimation of the time left, since not only the current I/O activity can affect the estimation but also a situation can occur in which some already scrubbed data might be removed during the scrubbing process. This would potentially result in a negative time left at the end of scrubbing, since the total bytes to scrub would decrease because of the removed space, while it would still be counted in the scrubbed bytes. I am not an expert, so maybe it is not completely true. Still, any estimate, even a rough one is more useful than no estimate at all.

BTW, the new output is indeed much more readable.

kdave:
I'll merge your patch as it works so you get the credit, and then add my that update the output and calculations.

PR #182, opened 2019-06-18
restore: Fix input buffer handling

pjw91:
`fgets` consumes **n-1** bytes from input buffer.

When a user types `y\n`, the newline is left in the buffer.
As a result, the next `fgets` uses that `\n` as answer without waiting for the user to type. 

This patch also fix a bug that dereference the `ret` without checking if it's `NULL`.


kdave:
Added to devel, thanks.

PR #188, opened 2019-06-28
restore: fix "looping" prompts, add `--looping` for unattended restores

intelfx:


kdave:
The looping caused by bad input buffer handling has been fixed by 064341dca9ed3a21106b5532e59e5f08833825a3, I'm leaving this open for the new option.

kdave:
The looping has been removed in 5.13. If there's ever need and real example where the loop detection really helps we can revive that. Thanks.

PR #190, opened 2019-07-07
Scrub status fixes

gkowal:
ETA is calculated  in a wrong way. It should be just current time in seconds + sec_left, independently if the job was resumed or not.

Also status lines for resumed jobs have wrong alignment.


kdave:
Merged, thanks.

Issue #201, opened 2019-09-03
Convert documentation build to docbook5 (asciidoctor)


kdave:
Reported on the mailinglist, that docbook45 is not supported by asciidoctor. Currently the build fails with docbook5 (validation fails). Workaround is to use --skip-validation for XMLTO, but we want to fix that.

https://docbook.org/docs/howto/howto.html how to port to 5

xnox:
> https://docbook.org/docs/howto/howto.html how to port to 5

is irrelevant... as your source input is asciidoc file which didn't change. And using docbook5 backend of asciidoctor valid manpages are still getting generated.

kdave:
I see, so the title is confusing. Using docbook5 as asciidoctor backend fails at the validation step (xmlto). I was not able to find what exactly is wrong. The original docs are copied from git but they don't have v5 support yet either.

PR #203, opened 2019-09-10
asciidoctor v2.0.0 droped docbook45 backend.

xnox:
See https://github.com/asciidoctor/asciidoctor/releases/tag/v2.0.0

Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>


CyberShadow:
See also: https://github.com/kdave/btrfs-progs/issues/201

xnox:
ah! true.

xnox:
@CyberShadow but i don't understand.... DocBook v5 does not use DTBs and xmlto tries to validate one... which makes no sense. Are you thinking to fix xmlto to skip DTB validation for DocBook v5?

PR #204, opened 2019-09-19
doc: update btrfs man page to explain compression level support

dennisszhou:
As of 5.1, btrfs now supports compression levels for zstd. Let users
know about this in the man page.

Signed-off-by: Dennis Zhou <dennis@kernel.org>


kdave:
Added to devel, with minor update, thanks.

PR #217, opened 2019-10-24
Added `btrfs replace` example for replacing an online drive

unguul:
Added example for replacing an online drive in a healthy state with a bigger one.


kdave:
Added to devel, thanks. I did some minor formatting edits.

PR #220, opened 2019-11-13
Documentation: btrfs-subvolume: Clarify the default list sort order

marcosps:
By default, "btrfs subvolume list" will show the subvolumes sorted by
subovlume id, in ascending order. Make this explicit in subvolume man
page.

Signed-off-by: Marcos Paulo de Souza <mpdesouza@suse.com>


kdave:
Applied, thanks.

PR #229, opened 2020-01-04
btrfs-progs: cmds: Allow setting compression level per file

### LINKED ISSUE #184: Allow setting zstd compression level with btrfs defrag
Currently btrfs defrag -c zstd uses a fixed compression level. It would be useful to be able to set it manually. The same way we can in mount options. zstd:1-15.

Thanks.

qwerty123443:
This fixes #184


qwerty123443:
I should note, this is in conjunction with https://github.com/kdave/btrfs-devel/pull/6

L3P3:
What about this conflict?
Sorry for dumb question but on my debian bullseye machine, I cannot use this defrag -c___:<level> feature. How can I see when and where it is available?

kdave:
This namely depends on the kernel part, the remaining work is in #184.

PR #244, opened 2020-04-18
Added example on deleting a subvolume

unguul:
Perhaps a rather trivial example but I find it helpful to see usage examples. Maybe it will help others.


kdave:
Absolutely, we need more examples, thanks.

PR #284, opened 2020-08-14
Fix typo in btrfs restore

gandalf3:
mirros -> mirrors


kdave:
Thanks, added to devel.

PR #287, opened 2020-08-23
Update btrfs-man5.asciidoc

Forza-tng:
Added filesystem limits: minimum filesystem size.


kdave:
Yes it would be better to make it clear when it depends on some other option.

Forza-tng:
> Yes it would be better to make it clear when it depends on some other option.

I updated the file. I hope I have not made a mistake with how git works. :)

kdave:
Merged to devel with some changes, thanks.

PR #301, opened 2020-09-22
btrfs-progs: Add examples in btrfs-qgroup documentation

realwakka:
This patch adds examples section in btrfs-qgroup documentation.
The example make a quota group that has two subvolumes as children.

Signed-off-by: Sidong Yang <realwakka@gmail.com>


realwakka:
@adam900710 Should I submit this with mailing list? I sent this PR cause there is a README that it's okay to send PR when updating docs.

adam900710:
Not sure, but I guess if the README says so, doc update PR should be no problem.

kdave:
Yes docs PRs are ok because there's usually nothing that would require others to see (via mails), also because documentation can be added by non-programmers so the PR should be easy way to contribute.
Merged to devel, thanks.

PR #321, opened 2020-12-07
Grammer and typo fix for fsck

esemeniuc:


kdave:
Merged to devel, thanks.

PR #331, opened 2021-01-20
fix wrong printf formats

neheb:
Signed-off-by: Rosen Penev <rosenp@gmail.com>


kdave:
So what I think should be changed is %Lu in place of %llu, and all the (unsigned long long) casts dropped unless necessary, and also type mismatches %lu instead of %zu.

PR #385, opened 2021-07-19
fix error for undefined macro on musl

wangmingyu84:
Fixes the following compilation errors with musl that does not have
NAME_MAX defined:
'NAME_MAX' undeclared (first use in this function)

Signed-off-by: Wang Mingyu <wangmy@cn.fujitsu.com>


kdave:
I lost track of this pull request that also sends a fix and created #386, now fixed in devel. Your fix works but NAME_MAX is defined in linux/limits.h and should be used as a file with the direct definition.

PR #401, opened 2021-09-14
btrfs-progs: check: add check for too many csum entries

maharmstone:
My tool ntfs2btrfs has been creating btrfs volumes in a way that the kernel doesn't like, but which isn't picked up by `btrfs check` - see https://github.com/maharmstone/ntfs2btrfs/issues/23 for the details, including a backtrace. This patch adds a check for when a csum item contains too many entries - effectively it ensures that there's always at least `sizeof(struct btrfs_item)` bytes free in the tree, otherwise `btrfs_del_csums` can throw an error.

`max_entries` is the value of the `__MAX_CSUM_ITEMS` macro in `fs/btrfs/file-item.c`.

### DIFF
--- a/check/main.c
+++ b/check/main.c
@@ -5987,6 +5987,10 @@ static int check_csums(struct btrfs_root *root)
 	u64 data_len;
 	unsigned long leaf_offset;
 	bool verify_csum = !!check_data_csum;
+	u16 num_entries, max_entries;
+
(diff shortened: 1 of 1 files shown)


kdave:
Thanks for the fix and sorry for late reply, I did not notice the pull request. Patch added to devel. As a side note, I'll mention the ntfs2btrfs tool in btrfs-convert page.

maharmstone:
Cheers Dave

PR #433, opened 2022-01-09
common/utils.c: drop include on sys/sysinfo.h

ffontaine:
Drop include on `sys/sysinfo.h` to avoid the following build failure on musl:

```
In file included from /usr/lfs/hdd_v1/rc-buildroot-test/scripts/instance-0/output-1/host/i586-buildroot-linux-musl/sysroot/usr/include/linux/kernel.h:4,
                 from ./kerncompat.h:31,
                 from common/utils.c:42:
/usr/lfs/hdd_v1/rc-buildroot-test/scripts/instance-0/output-1/host/i586-buildroot-linux-musl/sysroot/usr/include/linux/sysinfo.h:7:8: error: redefinition of 'struct sysinfo'
    7 | struct sysinfo {
      |        ^~~~~~~
In file included from common/utils.c:27:
/usr/lfs/hdd_v1/rc-buildroot-test/scripts/instance-0/output-1/host/i586-buildroot-linux-musl/sysroot/usr/include/sys/sysinfo.h:10:8: note: originally defined here
   10 | struct sysinfo {
      |        ^~~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/16f44fb9dea72a7079e8e5517e760dd79d2724cc

### DIFF
--- a/common/utils.c
+++ b/common/utils.c
@@ -24,7 +24,6 @@
 #include <sys/mount.h>
 #include <sys/types.h>
 #include <sys/stat.h>
-#include <sys/sysinfo.h>
 #include <uuid/uuid.h>
 #include <fcntl.h>
 #include <unistd.h>


kdave:
I verify builds on musl and haven't seen this error yet, is it something new? Or does it need some packages installed? https://github.com/kdave/btrfs-progs/blob/master/ci/images/ci-musl-x86_64/Dockerfile build from this.

kdave:
Removing the include produces this warning on non-musl build:

```
make[1]: Nothing to be done for 'all'.
common/utils.c: In function ‘total_memory’:
common/utils.c:1214:13: warning: implicit declaration of function ‘sysinfo’ [-Wimplicit-function-declaration]
 1214 |         if (sysinfo(&si) < 0) {
      | 
```

ffontaine:
linux-headers are installed.
Build failure is raised with musl on old "kernel" (i.e kernel without https://github.com/torvalds/linux/commit/a85cbe6159ffc973e5702f70a3bd5185f8f3c38d).

Would it be more acceptable to replace `<sys/sysinfo.h>` by `<linux/sysinfo.h>`?
If not, I'll keep this patch on buildroot side.

kdave:
`linux/sysinfo.h` only declares the structure, so the warning is still there as `sys/sysinfo.h` declares the function `sysinfo`. We could do some configure-time magic to detect it, or a simple "ifdef MUSL" for the include should work too.

kdave:
There don't seem to be an easy way to detect musl, so I'll go with the simplest fix and ifdef __GLIBC__ the include of `sys/sysinfo.h`, this is what `linux/kerne.h` does, this works on musl without explict inclusion of the sys path.

Can you please test this?

```
diff --git a/common/utils.c b/common/utils.c
index 29825f6ee15b..192ec2ffdb27 100644
--- a/common/utils.c
+++ b/common/utils.c
@@ -24,7 +24,9 @@
 #include <sys/mount.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#ifdef __GLIBC__
 #include <sys/sysinfo.h>
+#endif
 #include <uuid/uuid.h>
 #include <fcntl.h>
 #include <unistd.h>
```

kdave:
It's not that simple, now the CI build fails because the `sys/sysinfo.h` must be included, but it must not on your setup. This could be determined by a compilation check that does not produce the warnings about missing `sysinfo`.

kdave:
This seems to work on glibc build and CI/musl build. As `linux/const.h` is now included directly, the only other include is the sysinfo.h, so dropping the include here fixes the double definition, but please test it in your setup too.

```
--- a/kerncompat.h
+++ b/kerncompat.h
@@ -28,7 +28,6 @@
 #include <assert.h>
 #include <stddef.h>
 #include <linux/types.h>
-#include <linux/kernel.h>
 #include <linux/const.h>
 #include <stdint.h>
 ```

**Comment by ffontaine** [comment]
Commit b28f7bd9bba6f65258da48955154794f466b4104 and a2511aaa85c8d95d12805dfdcbb5667fa3e30ba5 are indeed fixing the issue. I'm closing this PR, thanks.

**Comment by ffontaine** [comment]
Unfortunately, both commits have the unexpected effect of raising a build failure with **kernel < 5.11** because `__ALIGN_KERNEL` is not defined anymore :

```
/tmp/instance-1/output-1/host/opt/ext-toolchain/bin/../lib/gcc/powerpc64-buildroot-linux-gnu/9.3.0/../../../../powerpc64-buildroot-linux-gnu/bin/ld: kernel-shared/volumes.o: in function `create_chunk':
volumes.c:(.text+0x17f8): undefined reference to `__ALIGN_KERNEL'
```

kdave:
That can be fixed by ifdefs and local definition if it's not defined but the order will matter, `kerncompat.h` is usually the first one while in `const.h` the macro `__ALIGN_KERNEL` is defined unconditionally.

kdave:
... or just copy & rename. Fix pushed to devel, please let me know if it works for you now.

ffontaine:
it works, thanks

PR #435, opened 2022-01-12
Fix incorrect statement

dathide:
Since free-space-tree is enabled by default in kernel 5.15, it is incorrect to state that no runtime options are enabled by default.


kdave:
I've applied the fix from #437, thanks.

PR #437, opened 2022-01-12
Remove incorrect statement

dathide:
Since 5.15 enables a runtime feature, this statement is incorrect.


kdave:
Applied to devel, thanks.

PR #442, opened 2022-01-29
docs: compression/zstd: direct maps note

jtagcat:
(info few months old) higher levels: upstream and library has levels 1..22.

https://btrfs.wiki.kernel.org/index.php?title=Talk:Compression&diff=0&oldid=33545

> mapping to the upstream levels 1 to 15

Where did you get that infobit? I'm taking your word for it.

Also, is readthedocs staged to replace the wiki at some point?


kdave:
>> mapping to the upstream levels 1 to 15

> Where did you get that infobit? I'm taking your word for it.

In the sources, https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/btrfs/zstd.c#n27

```c
#define ZSTD_BTRFS_MAX_LEVEL 15
```

And there's no translation between the user requested and zstd levels. Initial workspaces are created in 
zstd_calc_ws_mem_sizes for each level and then per request the one available is selected.

> Also, is readthedocs staged to replace the wiki at some point?

Yes, though people are used to it and there are random pages linking to it so it won't go away completely. For contents that's not suitable for project documentation the wiki will be a good place, like the news or some community links.

kdave:
Patch applied to devel with a slightly updated changelog, thanks.

PR #443, opened 2022-02-05
mkfs/main.c: fix build without zoned

ffontaine:
Fix the following build failure raised when zoned is disabled since version 5.16.1 and https://github.com/kdave/btrfs-progs/commit/88895a920ff5b0969ec22fdc9f4511b17dd14806:

```
mkfs/main.o: In function `main':
main.c:(.text.startup+0xc90): undefined reference to `zoned_profile_supported'
```

### DIFF
--- a/mkfs/main.c
+++ b/mkfs/main.c
@@ -1349,11 +1349,13 @@ int BOX_MAIN(mkfs)(int argc, char **argv)
 	if (ret)
 		goto error;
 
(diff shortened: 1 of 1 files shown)


kdave:
Thanks, it's been fixed in a different way via #445. There should be a stub for the zoned related helpers in the respective sections in kernel-shared/zoned.h, instead of the inline ifdefs.

PR #454, opened 2022-03-16
btrfs-progs: prop: add datacow inode property

martinetd:
Hello.

```
The btrfs property documentation states that it is an unified and
user-friendly method to tune btrfs properties instead of chattr,
so let's add something for datacow as well.
```

A few remarks:
 - please tell me if you prefer this sent on linux-btrfs@vger list, I don't mind either workflow but I saw other open PRs here and you seem to use github issues at least.
 - naming: I wasn't sure whether to name it datacow with yes/no, or making it "nodatacow" with true/false (readonly uses true/false so it might make more sense to use the later), I've picked datacow to avoid double-negation for ease of understanding but happy to change it to anything
 - documentation: I got a bit confused with the rst and asciidoc file, as things got "converted" to rst recently but the asciidoc file didn't get removed. Should I have updated both?

Thanks!

### DIFF
--- a/Documentation/btrfs-man5.asciidoc
+++ b/Documentation/btrfs-man5.asciidoc
@@ -712,7 +712,7 @@ To create and activate a swapfile run the following commands:
 
 --------------------
 # truncate -s 0 swapfile
-# chattr +C swapfile
+# btrfs property set swapfile datacow no
 # fallocate -l 2G swapfile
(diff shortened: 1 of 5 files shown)


martinetd:
Continuing on mailing list: https://lore.kernel.org/all/20220324042235.1483914-1-asmadeus@codewreck.org/T/#u

PR #456, opened 2022-03-16
btrfs-progs: check: add check for overlong xattr names

maharmstone:
While working on my Windows driver, I found that it was inadvertently allowing users to create xattrs with names longer than 255 bytes, which wasn't being picked up by btrfs-check.

If the Linux driver encounters a file with an invalid xattr like this, it makes the whole directory it's in inaccessible. If it's the root directory, it'll refuse to mount the filesystem entirely.


kdave:
Applied to devel, thanks.

PR #462, opened 2022-04-14
discourage seeding remount workflow

boryas:
Don't want to erase it entirely from the doc, since it is the ideal way to use it ultimately. However, the bug is significant, so I think it's our responsibility to encourage users that can use umount/mount (e.g. in an initrd, not for root fs) to use it.


kdave:
Merged to devel with some minor edits. The description of the workflows can be still improved, separating clearly the use case without snapshots that the remount is fine.

PR #469, opened 2022-05-19
Clarification on mixed profile

Forza-tng:
It is not possible to use mixed profile together with other profiles. The current wording is not clear about this, so let's add a clarification note.


kdave:
Merged, thanks.

PR #474, opened 2022-05-24
btrfs-progs: add option to disable LZO support

rossburton:
LZO as a compression format is pretty archaic these days, there are
better algorithsm in all metrics for compression and decompression, and
lzo hasn't had a new release since 2017.

Add an option to disable LZO (defaulting to enabled), and respect it in
cmds/restore.c.

Signed-off-by: Ross Burton <ross.burton@arm.com>


kdave:
I don't mind adding it but I don't think it's a good idea to remove support for the compression namely for the restore command as it's a rescue tool and that should cover more rather then less of possible features. I understand that minimizing dependencies is a good thing or that in controlled environments it makes sense to remove what's not used. For a generic distro I'm strongly against using the option.

rossburton:
Sure, that's why this is an opt-in disable, I don't expect you to remove it entirely.

PR #478, opened 2022-05-29
make description of list output more clear

frukto:


kdave:
Added to devel, thanks.

PR #480, opened 2022-06-16
btrfs-progs: docs: rm reflink cross-mount constraint

AtticFinder65536:
I just noticed it was still there whilst reading now that it works in 5.18


kdave:
Thanks, I've kept the paragraph but mentioned the version where it started to work. People are using older versions so it would be confusing to remove it completely. Merged to devel.

PR #530, opened 2022-10-14
Update ch-swapfile.rst

createyourpersonalaccount:
The fact that the +C attribute excludes compression is mentioned in <https://btrfs.readthedocs.io/en/latest/ch-compression.html#compatibility>.


kdave:
Merged to devel, thanks.

PR #563, opened 2023-01-01
Static builds should pick up EXTRA_LDFLAGS and SUBST_LDFLAGS

snarkmaster:
Concrete use-case:
 - I want a static build to run in a small test VM.
 - RedHat regrettably does not provide `util-linux` static libraries like `libuuid.a` and `libblkid.a`. 
 - So, one has to build them from source.
 - Installing these to `/usr` is lame.
 - This change lets me do `make btrfs.static EXTRA_LDFLAGS=-L/path/to/util-linux-2.38/.libs`


kdave:
Merged to devel, thanks. Missing EXTRA_LDFLAGS from static build was not intentional, the option is meant to be used for build tweaks like that.

PR #565, opened 2023-01-03
Update ch-mount-options.rst :: Add main title

pinysuse:
The header of the main section seems to be missing --> added.

This might fix the strange appearance on [readthedocs](https://btrfs.readthedocs.io)

Screenshot_20230103_194341

where all those options appear to be deprecated!


kdave:
Added to devel, thanks.

PR #569, opened 2023-01-15
Document some ioctls

Hawk777:


kdave:
Added to devel with some minor edits to changelogs, thanks.

PR #571, opened 2023-01-16
btrfs-progs: docs: improve space cache documentation

calestyo:
- "Wipe" in storage terms is often understood as some kind of secure deletion. Use "remove" instead in order to indicate that the space cache is fully removed (and not just cleared and then e.g. automatically rebuild).

- The --clear-space-cache option for btrfs check actually clears the whole space cache, just as documented. Thus move any documentation about the clear_cache mount option not doing so for v1 to that. Instead, refer to the mount option.

- Also note that when clear_cache is used with v1, the free space cache for block groups that are modified gets always cleared, but rebuilt only if nospace_cache is not used.

Signed-off-by: Christoph Anton Mitterer <mail@christoph.anton.mitterer.name>

### DIFF
--- a/Documentation/btrfs-check.rst
+++ b/Documentation/btrfs-check.rst
@@ -79,17 +79,9 @@ SAFE OR ADVISORY OPTIONS
         superblock is damaged.
 
 --clear-space-cache v1|v2
-        completely wipe all free space cache of given type
+        completely remove the free space cache of the given version
 
-        For free space cache *v1*, the *clear_cache* kernel mount option only rebuilds
-        the free space cache for block groups that are modified while the filesystem is
-        mounted with that option. Thus, using this option with *v1* makes it possible
-        to actually clear the entire free space cache.
-
-        For free space cache *v2*, the *clear_cache* kernel mount option destroys
-        the entire free space cache. This option, with *v2* provides an alternative
-        method of clearing the free space cache that doesn't require mounting the
-        filesystem.
+        See also the *clear_cache* mount option.
 
 --clear-ino-cache
         remove leftover items pertaining to the deprecated inode map feature
--- a/Documentation/ch-mount-options.rst
+++ b/Documentation/ch-mount-options.rst
@@ -89,8 +89,19 @@ check_int, check_int_data, check_int_print_mask=<value>
         for more information.
 
 clear_cache
-        Force clearing and rebuilding of the disk space cache if something
-        has gone wrong. See also: *space_cache*.
+        Force clearing and rebuilding of the free space cache if something
+        has gone wrong.
+
+        For free space cache *v1*, this only clears (and, unless *nospace_cache* is
+        used, rebuilds) the free space cache for block groups that are modified while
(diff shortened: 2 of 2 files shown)


calestyo:
See open questions at:
https://lore.kernel.org/linux-btrfs/c7bf7a6053211c1ac84ab00441e7505df25cbbe6.camel@scientia.org/T/#u

calestyo:
@kdave have you had a chance to look at this?

calestyo:
(rebased on master)

kdave:
"a) Does --clear-space-cache v1|v2 really remove (or just clear) any
   space cache from the fs? Or is there even a difference?
 
   Especially, when I ran --clear-space-cache v1|v2 and then mount the
   fs (without neither nospace_cache nor space_cache[=*])...
   will it automatically get a new space cache, if so, which?"

--clear-space-cache option really deletes all the space structures, i.e. the free space inode or the free space tree. After that mount will pick v2 if the free-space-tree compat bit is set, otherwise v1.

kdave:
"b) Is my assumption correct, that with v1, clear_cache itself really
   only clears the space cache for written blocks but not automatically
   rebuilds it.
   I.e. when clear_cache,nospace_cache would be used, its just cleared
   but not rebuild."

Clear_cache resets the state and allows rebuilding, with nospace_cache the caching is disabled, so when both are used it is a you say. I hope I got it right, it's a maze of various flags and transitions in the state machine of the cache rebuild, but to my best understanding your assumptions are correct.

kdave:
"c) What about clear_cache and v2? Is it only clearing? Or is it,
   unless nospace_cache is used, right after clearing completely
   rebuilt?"

Same as for v1, i.e. clear_cache first deletes the tree and later in mount process the free-space-tree feature is checked, if set the tree is rebuilt. With nospace_cache rebuilding and caching would be skipped.

kdave:
The changes in this PR seem to be in line with that so I'll apply it. I'm about to release 6.3 so if you have further updates please send it separately. Thanks.

calestyo:
Uhm... sorry but I think there is still some mess in here:

a)
> --clear-space-cache option really deletes all the space structures, i.e. the free space inode or the free space tree. After that mount will pick v2 if the free-space-tree compat bit is set, otherwise v1.

Okay, so
https://github.com/kdave/btrfs-progs/blob/405f4c73e8931235df242ff6bb83246f104c049e/Documentation/btrfs-check.rst?plain=1#L81-L82
is IMO fine.
But the part, that on next mount a new space cache will be generated was missing - well not strictly speaking but I think it would be good to have it mentioned for convenience.

Also, moving the stuff about `clear_cache` from `btrfs-check.rst` to `ch-mount-options.rst` is fine.


b) and c)
It wasn't really mentioned for v2 in my old commit, that a rebuild would happen "immediate", also not that for v2 the `nospace_cache` would prevent it, too.

Also made some further minor improvements (IMO) to structure.

I've made PR #620 with these changes.

PR #587, opened 2023-03-02
Makefile: guard x86-specific acceleration with an explicit option

kanavin:
Using 'uname -m' to make this decision really does not work when cross compiling (e.g. for an arm target on an x86 host), so this feature needs to be opt-in.


kdave:
I've fixed it by dropping the uname check and it builds fine on i586 (4fc291a46516a46c1750ce01d212ccbe195d72cf), making a config option is not necessary as it picks the support from complier automatically. Bugfix release 6.2.1 will be out soon, but thanks anyway.

Issue #589, opened 2023-03-04
blake2b-sse2: incompatible type for argument


heitbaum:
With both 6.2 and 6.2.1 the following blake2b-sse2 - incompatible type for argument build errors.

```
docker@280b00be4b8d:/var/media/DATA/home-rudi/LibreELEC.tv$ s/build btrfs-progs
UNPACK      btrfs-progs
BUILD      btrfs-progs (target)
    TOOLCHAIN      configure

Generate build-system by:
   aclocal:    aclocal (GNU automake) 1.16.5
   autoconf:   autoconf (GNU Autoconf) 2.71
   autoheader: autoheader (GNU Autoconf) 2.71
   automake:   automake (GNU automake) 1.16.5

Now type './configure' and 'make' to compile.

Executing (target): /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-11.0-devel/build/btrfs-progs-6.2.1/configure --host=x86_64-libreelec-linux-gnu --build=x86_64-linux-gnu --prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc --libdir=/usr/lib --libexecdir=/usr/lib --localstatedir=/var --disable-static --enable-shared --disable-backtrace --disable-convert --disable-documentation --disable-python --disable-zstd
checking for x86_64-libreelec-linux-gnu-gcc... /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-11.0-devel/toolchain/bin/x86_64-libreelec-linux-gnu-gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... yes
checking for suffix of object files... o
checking whether the compiler supports GNU C... yes
checking whether /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-11.0-devel/toolchain/bin/x86_64-libreelec-linux-gnu-gcc accepts -g... yes
checking for /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-11.0-devel/toolchain/bin/x86_64-libreelec-linux-gnu-gcc option to enable C11 features... none needed
checking for stdio.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for strings.h... yes
[...]

PR #591, opened 2023-03-05
btrfs-progs: crypto: fix SSE2/SSE4.1 implementation of BLAKE2

tinomai:
On recent x86-64 system with `march=native|<cpu>|<microarchitecture level>` gcc/clang will automatically define all the available vector extensions macros. `crypto/blake2-config.h` then correctly set all the `HAVE_<EXTENSION>` macros. 

`crypto/blake2-round.h` then checks the `HAVE_<EXTENSION>` macros for including further headers:

>#if defined(HAVE_SSE41)
#include "blake2b-load-sse41.h"
#else
#include "blake2b-load-sse2.h"
#endif

which is wrong. On recent systems it always results in including `blake2b-load-sse41.h`. `crypto/blake2-round.h`  itself is included by `crypto/blake2b-sse2.c` and now we have a SSE2/SSE4.1 code mixing resulting in the incompatible type for argument build errors described in #589.

The idea is to remove the lines above from `crypto/blake2-round.h` and put the includes directly into `crypto/blake2b-sse2.c` and `crypto/blake2b-sse41.c` respectively.

### DIFF
--- a/crypto/blake2b-round.h
+++ b/crypto/blake2b-round.h
@@ -136,12 +136,6 @@
 
 #endif
 
-#if defined(HAVE_SSE41)
-#include "blake2b-load-sse41.h"
-#else
-#include "blake2b-load-sse2.h"
-#endif
-
 #define ROUND(r) \
   LOAD_MSG_ ##r ##_1(b0, b1); \
   G1(row1l,row2l,row3l,row4l,row1h,row2h,row3h,row4h,b0,b1); \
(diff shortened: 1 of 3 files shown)


kdave:
Thanks, the fix slightly diverges from the upstream sources but I'll add some build check should the code get synced in the future.

kdave:
Merged to devel, thanks.

PR #596, opened 2023-03-14
btrfs-progs: docs: fix name of "mountinfo" file in subvolume intro

alexbarton:
The name of a mounted sub volume can not be found in `/proc/self/mounts` but in `/proc/self/mountinfo`.


kdave:
Merged to devel, thanks.

PR #599, opened 2023-03-15
btrfs-progs: handle case if fsid not in sysfs in device_is_seed

olifre:
For older kernels, the sysfs interface providing the fsid may not be present yet. Since 32c2e57c65b9974c5ec709a6bf47d65040bdf9a1, the fallback to the previous approach to determine the fsid was not used anymore.

Ensure negative return values of `sysfs_open_fsid_file` are handled by falling back to the `dev_to_fsid` in this case.


kdave:
Merged to devel, thanks.

PR #601, opened 2023-03-16
Update ch-swapfile.rst Added size to mkswapfile example

softwarecreations:
Thanks for maintaining these docs, they were very helpful!

Let's show users how to set the size of their new swapfile.


kdave:
Thanks, showing the size option in the example makes sense, the default size is there as a sane fallback. I've moved the option before the swapfile name so it's following the command/options/non-opts convention.

softwarecreations:
oh, that's great, thanks!

PR #615, opened 2023-04-15
btrfs-progs: stop use of legacy *64 interfaces

nekopsykose:
```
the *64 interfaces, such as fstat64, off64_t, etc, are legacy interfaces       
created at a time when 64-bit file support was still new. they are             
generally exposed when defining a macro named _LARGEFILE64_SOURCE, as          
e.g. the glibc docs[0] say.                                                    
                                                                               
the modern way to utilise largefile support, is to continue to use the         
regular interfaces (off_t, fstat, ..), and define _FILE_OFFSET_BITS=64.        
                                                                               
we already use the autoconf macro AC_SYS_LARGEFILE[1] which arranges this      
and sets this macro for us. therefore, we can utilise the non-64 names         
without fear of breaking on 32-bit systems.                                    
                                                                               
this fixes the build against musl libc, ever since musl dropped the            
*64 compat from interfaces by default[2] just for _GNU_SOURCE, unless          
_LARGEFILE64_SOURCE is defined. however, there are plans for a future          
removal of the whole *64 header api, and that workaround (adding another       
define) might cease to exist.                                                  
                                                                               
so, rename all *64 api use to the regular non-suffixed names. for              
consistency, rename the internal functions that were *64 named                 
(lstat64_path, ..) too.                                                        
                                                                               
this should have no regressions on any platform.          

[0]: https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html#index-_005fLARGEFILE64_005fSOURCE
[1]: https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/System-Services.html
[2]: https://github.com/bminor/musl/commit/25e6fee27f4a293728dd15b659170e7b9c7db9bc
```

### DIFF
--- a/cmds/rescue-chunk-recover.c
+++ b/cmds/rescue-chunk-recover.c
@@ -754,7 +754,7 @@ static int scan_one_device(void *dev_scan_struct)
 		if (is_super_block_address(bytenr))
 			bytenr += rc->sectorsize;
 
-		if (pread64(fd, buf->data, rc->nodesize, bytenr) <
+		if (pread(fd, buf->data, rc->nodesize, bytenr) <
 		    rc->nodesize)
 			break;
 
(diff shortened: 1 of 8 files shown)


nekopsykose:
i'm not sure if it would be wanted or not to actually keep the `printf` naming (e.g. `"fsstress: check_cwd stat64() returned %d with errno: %d (%m)\n"`) untouched.

nekopsykose:
thank you!

PR #620, opened 2023-04-26
btrfs-progs: docs: further improve space cache documentation

calestyo:
- For the btrfs check option of clearing, indicate that - since it's actually a full removing - a new space cache may be generated on the next mount.

- In the summary of the `clear_cache` mount option, don't mention that it rebuilds the cache, as the option itself is in principle only about the clearing and the rebuilding happens "automatically" (unless other options are used).

- In the details mention, however that there is rebuilding unless prevented by mount option.

- Remove the "if something has gone wrong", as the mount option would also remove the cache if nothing went wrong.

### DIFF
--- a/Documentation/btrfs-check.rst
+++ b/Documentation/btrfs-check.rst
@@ -81,6 +81,8 @@ SAFE OR ADVISORY OPTIONS
 --clear-space-cache v1|v2
         completely remove the free space cache of the given version
 
+        When the filesystem is mounted afterwards, a new space cache may be generated
+        depending on mount options and/or kernel defaults.
         See also the *clear_cache* mount option.
 
 --clear-ino-cache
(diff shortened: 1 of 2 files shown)


calestyo:
please wait with this one... there's more to come

calestyo:
Correctly rebased on `devel`... but there's one question open for the 2nd commit:

https://github.com/kdave/btrfs-progs/blob/e8a471c5e7dad388e1dda7c01fe46e89baef3ed4/Documentation/ch-mount-options.rst?plain=1#L364

Is the old v1 then cleared or (as with `btrfs check --clear-space-cache v2`) completely removed (if the latter, the wording should be adapted).

adam900710:
The PR is pretty old, and we are moving to completely removing v1 cache in kernel soon.

If you still want to update the docs, please open a new PR.

PR #622, opened 2023-05-01
btrfs-progs: check: check order of inline extent refs

maharmstone:
The kernel seems to order inline extent items in a particular way: forward by sub-type, then reverse by hash. Having these out of order can cause a volume to go readonly when deleting an inode.

See https://github.com/maharmstone/ntfs2btrfs/issues/51

### DIFF
--- a/check/main.c
+++ b/check/main.c
@@ -5546,13 +5546,14 @@ static int process_extent_item(struct btrfs_root *root,
 	unsigned long end;
 	unsigned long ptr;
 	int ret;
-	int type;
+	int type, last_type;
 	u32 item_size = btrfs_item_size(eb, slot);
 	u64 refs = 0;
 	u64 offset;
 	u64 num_bytes;
 	u64 gen;
 	int metadata = 0;
+	u64 last_hash, hash;
 
 	btrfs_item_key_to_cpu(eb, &key, slot);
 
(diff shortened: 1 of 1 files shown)


kdave:
@adam900710 Qu can you please have a look?

adam900710:
@maharmstone  Do you still have such bad fs or could you craft one?
We'd like to have a test case for such case.

The patch itself looks goot to me though.

Thanks for exposing the bug!

adam900710:
I checked the kernel code, and indeed `lookup_inline_extent_backref()` is skipping the remaining backref item if data/metadata item is smaller (either through the data hash, or metadata parent/ref_root) than the target range.

Thus it indeed requires the full order inside the inline extent backref.

But we still need the following two things:

- Also detect such problem in lowmem mode

- A test case for it

  For test case, do you still have such image?
  If not, I can help to craft such image if you don't have spare time.
  The same for lowmem mode detection support.

maharmstone:
@adam900710, have a look at @sleirsgoevy's comment in the issue I linked, which gives you the steps to create a test case. Make sure you use ntfs2btrfs 20220812, as it's been fixed since.

kdave:
I've merged the pull request. We don't have the test case and as said above lowmem mode is probably not fixed, but I'd like to move forward with this PR. I've updated the changelog with remaining issues so it's not completely forgotten.

maharmstone:
Thanks Dave

adam900710:
I'll handle the lowmem mode part along with the image part.

Great work exposing the hidden restrict @maharmstone .

For the image part, I guess I had to go some custom `btrfs-corrupt-block` to create a minimal image.
Overall it shouldn't be that hard.

Maybe I can even add tree-checker for the problem, so that the next time we hit such situation, kernel would give a more detailed output to expose such corrupted fs.

adam900710:
Follow ups for progs are here:
https://lore.kernel.org/linux-btrfs/cover.1697945679.git.wqu@suse.com/

The image is hand crafted with 3 inline data backref items, and switched the firsst 2.

kdave:
Thanks, fix added to devel.

PR #623, opened 2023-05-01
btrfs-progs: check: add check for metadata level mismatch

maharmstone:
For a skinny metadata item in the extent tree, the key offset represents the level of the tree it points to. This adds a check that these values match, as otherwise it can cause a volume to go readonly when deleting a large number of inodes.

See https://github.com/maharmstone/ntfs2btrfs/issues/51


kdave:
@adam900710 Qu can you please have a look?

adam900710:
The patch looks good to me.
`Reviewed-by: Qu Wenruo <wqu@suse.com>`

And thankfully lowmem mode is already detecting such problem, so no need for a lowmem mode fix.

But we still want a test case, I can craft one if needed.

kdave:
Thanks, added to devel. The test case would be great.

PR #633, opened 2023-05-29
mkfs: make --quiet silence the 5.15 default change NOTE

martinetd:
mkfs.btrfs help message for --quiet is 'no message except errors' so we probably ought to silence this as well in the quiet case.

-----

It's easy enough to redirect stdout to /dev/null but might as well have --quiet do what it says.
(feel free to consider this a non-issue if this is on purpose, but there wasn't any other issue or PR about this -- I'll admit I didn't check the ML)

Thanks!


martinetd:
Thanks!

PR #635, opened 2023-06-03
check: add a specific warning in --help for repair

thesamesam:
When running 'btrfs check --repair', we get a nice runtime warning and a pause for 10 seconds, but there's no indication of the risks in --help (or no explicit one, anyway). Add it.


kdave:
Merged to devel, thanks.

PR #675, opened 2023-09-05
btrfs-progs: crypto: fix readonly relocation of the jumptable

kaniini:
Without this, the btrfs programs fail to link when build with -Wl,-z,relro, due to the jumptable containing relocations.


nekopsykose:
i tested removing the GLIBC ifdefs from 992be8b50a65974e52a484911e72bc5a5e1b412c, and this now works fine on musl , so you could also lift those i think

kdave:
Thanks for the fix. I've removed the glibc workaround as a separate patch as well, thanks for the suggestion.

kdave:
The 32bit build on musl fails because the asm implementation is using 64bit instructions. The glibc 32bit works in OBS but would probably fail to run on a real 32bit machine.

PR #700, opened 2023-10-29
typo in btrfs-rescue.rst

Martinligabue:
that are do not match -> that do not match


kdave:
Merged to devel, thanks.

PR #707, opened 2023-11-06
Fix -Walloc-size

thesamesam:
GCC 14 introduces a new -Walloc-size included in -Wextra which gives:
```
common/utils.c:983:15: warning: allocation of insufficient size ‘1’ for type ‘struct config_param’ with size ‘32’ [-Walloc-size]
cmds/qgroup.c:1644:13: warning: allocation of insufficient size ‘1’ for type ‘struct btrfs_qgroup_inherit’ with size ‘72’ [-Walloc-size]
```

The calloc prototype is:
```
void *calloc(size_t nmemb, size_t size);
```

So, just swap the number of members and size arguments to match the prototype, as we're initialising 1 struct of size `sizeof(struct ...)`. GCC then sees we're not doing anything wrong.

### DIFF
--- a/cmds/qgroup.c
+++ b/cmds/qgroup.c
@@ -1641,7 +1641,7 @@ static int qgroup_inherit_realloc(struct btrfs_qgroup_inherit **inherit, int n,
 			 (*inherit)->num_excl_copies;
 	}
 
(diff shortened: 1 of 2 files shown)


thesamesam:
Thank you!

PR #721, opened 2023-12-15
configure.ac: fix bashism

thesamesam:


kdave:
Added to devel, thanks.

PR #740, opened 2024-02-07
Fix tiny typo in README.md

moontide:


kdave:
Applied to devel, thanks.

PR #744, opened 2024-02-16
btrfs-progs: docs: Correct systemd-run argument for limiting bandwith for scrub

Termuellinator:
Hi, i've stumbled upon a small error in the documentation - the argument is `IOReadBandwidthMax` instead of `IOBandwidthReadMax`. 
See https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html


kdave:
Added to devel, thanks.

PR #754, opened 2024-02-29
Docs/conf.py: Set navigation_with_keys to allow navigating documentation through arrow keys

C0rn3j:
Feature from https://github.com/sphinx-doc/sphinx/pull/2064.

I have not tested this (not sure how I would), but judging from other readthedocsprojects, it seems to be the correct way to set enable this.


kdave:
Sounds useful, thanks. We'll see if it works in a minute.

kdave:
Left/right keys work.

C0rn3j:
That was speedy, thanks a lot!

PR #759, opened 2024-03-18
Typo in ch-hardware-considerations.rst

tbertels:
verity -> verify


kdave:
Thanks, that's what spellchecked does not catch. Added to devel.

PR #762, opened 2024-03-21
Some user-permission changes to btrfs-progs

sweettea:
This makes it possible to use btrfs receive as a regular user, and gets part of the way to using btrfs subvolume list as a regular user (the main change there is pretty big so it's going in a different pull request).

### DIFF
--- a/cmds/receive.c
+++ b/cmds/receive.c
@@ -1285,7 +1285,7 @@ static int process_encoded_write(const char *path, const void *data, u64 offset,
 		if (ret >= 0)
 			return 0;
 		/* Fall back for these errors, fail hard for anything else. */
-		if (errno != ENOSPC && errno != ENOTTY && errno != EINVAL) {
+		if (errno != ENOSPC && errno != ENOTTY && errno != EINVAL && errno != EPERM) {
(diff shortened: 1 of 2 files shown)


adam900710:
For data writes it's totally fine to fallback to non-privileged writes.

But wouldn't things like snapshotting and UUID tree updates still requires root privilege?

sweettea:
Snapshotting shouldn't require root?

But uuid search, good point, I was running on a kernel with uuid search allowed for non-root.

kdave:
There are the unprivileged subvolume ioctls, I'm not sure if there are also helpers for that but this should be possible to implement by manual iteration.

kdave:
I'm taking another look, the idea of using unprivileged ioctls makes sense but we can go further and use the library functions (receive code is old so the interface was not there). I'd apply the patches but the change in lookup_ino_path() to `strcpy(ri->name, args.name);` looks like it's ignoring the following code that checks `args.name[0]` and build the path.

adam900710:
Is the PR still alive? Although it still applies I haven't seen any update for 2 years.

If you're still interested in pushing this feature, please open a new PR.

PR #764, opened 2024-03-23
btrfs-progs: docs: fix incorrect description about compression with O…

hanyuwei70:
According to the mail below, I found that docs about compression is incorrect.
After reading source code, I corrected these descriptions.

http://lore.kernel.org/linux-btrfs/e7ce9995-93cb-4904-875c-684d4494765f@web.de/


kdave:
Right, thanks.

PR #766, opened 2024-03-26
Zoned bgt

adam900710:
There is a bug report that, the following tool would fail on zone
devices:

- mkfs.btrfs -O block-group-tree
- btrfstune --convert-to-block-group-tree|--convert-from-block-group-tree

The mkfs failure is caused by zoned incompatible pwrite() calls for
block group tree metadata.

The btrfstune failure is caused by the incorrectly opened fd.

Before fixing both bugs, do two small cleanups, one caught by clangd LSP
server, the other caught by my later check on the test case output (a
missing newline).

Then fixes for each bug, and new test cases for each bug.


adam900710:
Wrong target branch.

PR #767, opened 2024-03-26
Zoned bgt

adam900710:
There is a bug report that, the following tool would fail on zone
devices:

- mkfs.btrfs -O block-group-tree
- btrfstune --convert-to-block-group-tree|--convert-from-block-group-tree

The mkfs failure is caused by zoned incompatible pwrite() calls for
block group tree metadata.

The btrfstune failure is caused by the incorrectly opened fd.

Before fixing both bugs, do two small cleanups, one caught by clangd LSP
server, the other caught by my later check on the test case output (a
missing newline).

Then fixes for each bug, and new test cases for each bug.

### DIFF
--- a/mkfs/common.c
+++ b/mkfs/common.c
@@ -249,8 +249,8 @@ static int create_block_group_tree(int fd, struct btrfs_mkfs_config *cfg,
 	btrfs_set_header_nritems(buf, 1);
 	csum_tree_block_size(buf, btrfs_csum_type_size(cfg->csum_type), 0,
 			     cfg->csum_type);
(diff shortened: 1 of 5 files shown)


oxalica:
Thanks for the fix. I built btrfs-progs with this patch applied, and it (`btrfstune`) successfully converted my filesystem on network-based block device to be with bgt enabled. It mounts without any issues and the mounting time reduced from ~50s to ~5s, probably because of the high latency of the backing device.
The new `mkfs.btrfs` also works now on my emulated block device.

kdave:
Thanks for the fix and testing. It's been in devel for some time, now closing the issue.

PR #769, opened 2024-03-31
Headers cleanup

adam900710:
This series is focusing on cleanup the unused headers.

This is mostly done by clangd, although it has some false alerts related
to macro usages of a header.

But still it's pretty awesome to cleanup a lot of unnecessary headers.
Only one special touch on pretty_size() macro, to change it to a static
inline function to workaround the clangd bug.

The first patch would do the main heavy lifting, meanwhile the second
patch is doing the BTRFS_FLAT_INCLUDES related cleanups for library-test
code.

Unfortunately I didn't touch anything inside crypto/*, the main reason
is I'm not confident enough to verify all the optimization for different
instructions.

(And even less motivation after the infamous recent XZ backdoor attempt,
 by a possibly state-sponsored sleeping agent, ruining the trust among
 open source community.)

### DIFF
--- a/btrfs-corrupt-block.c
+++ b/btrfs-corrupt-block.c
@@ -20,7 +20,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <getopt.h>
-#include <limits.h>
 #include <errno.h>
 #include <string.h>
 #include <unistd.h>
--- a/btrfs-sb-mod.c
+++ b/btrfs-sb-mod.c
@@ -20,11 +20,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
-#include <inttypes.h>
 #include <string.h>
 #include <limits.h>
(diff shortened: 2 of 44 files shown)


kdave:
IIRC the limits.h must stay otherwise PATH_MAX is not found in some builds but I'd have to verify that. The `box.h` is used conditionally for the  `make box` target, maybe this is missing from the build tests. Please keep `kerncompat.h` where it is.

adam900710:
Any CI or selftest that can iterate all the configure options?
Indeed the LSP server is a little too aggressive, but I'd like the CI system to give me a big failure earlier.

For now, please drop this series.

kdave:
CI now covers the btrfs.box target and experimental build.

kdave:
For the LSP it's possible to generate the `compile_commands.json` file that `clangd` uses but this would still need additional logic above that to determine if a symbol or include is used in any of the configs.

PR #771, opened 2024-04-02
fix image directives in btrfs-design page

austin880625:
Copy the images from wiki so that we don't need to jump around the web search results


kdave:
Added to devel, thanks.

PR #792, opened 2024-05-09
btrfs-progs: cmds/qgroup: enhance stale qgroups handling

adam900710:
Currently the `<stale>` label is give incorrectly for a lot of qgroups
that can not and should not be deleted.

The patchset would address the problem for `btrfs qgroup show` and
`btrfs qgroup clear-stale` commands, so that:

- `clear-stale` would not try to delete incorrect qgroups
  Thus it should not return false errors.

- `show` would show extra types of qgroups
  Including `<under deletion>` and `<squota space holder>`


kdave:
Added to devel, thanks.

PR #793, opened 2024-05-10
Update CHANGES

IlgazC:
Fixed typos on line 7 and 9.


IlgazC:
While they are typos, they look like being written from a severely lagging computer or broken keyboard.

kdave:
Added to devel, thanks.

PR #800, opened 2024-05-31
Disable faulty check for a running dev-replace while `btrfstune --csum`ing

### LINKED ISSUE #798: Changing Checksum Fails "running dev-replace detected"
```
sudo ./btrfstune --csum xxhash64 /dev/sdd
Proceeding to switch checksums
ERROR: running dev-replace detected, please finish or cancel it.
ERROR: btrfstune failed
```
There is no dev-replace running, although I have replaced devices on this fs before.

Note: This fs is actually two drives using RAID 0
```
sudo btrfs fi show /dev/sdd
Label: 'Data11' uuid: blah
Total devices 2 FS bytes used 28.46 TiB
devid 1 size 10.91TiB used 10.91TiB path /dev/sdc1
devid 3 size 18.19TiB used 18.17TiB path /dev/sdd
```

I also tried this a while back, about a year ago I believe, and got the same error. Maybe I botched the dev-replaces in the past? How can I troubleshoot further?

I have already run `sudo btrfs check --check-data-csum --progress /dev/sdd` without errors (took several days to run). 

I am using the latest btrfs-progs, built from git. I have also successfully changed to xxhash64 on my root drives, also RAID1, on the same machine. It really seems like there is something messed up in the metadata (or something) for Data11. 

Very willing to do the legwork on this, just tell me what to do. Thanks!

aidan-gibson:
This code I'm commenting out doesn't check for a *running* dev-replace, as intended. It checks if a dev-replace has ever been done. I'm unsure if the intention of `BTRFS_DEV_REPLACE_KEY` was to monitor currently a currently running dev-replace and wasn't implemented properly, or it was intended to keep a history of dev-replaces and is functioning properly, but as this code is currently written, you cannot `btrfstune --csum` if a dev-replace has ever been done. 

Tested on Arch and Debian to be sure.

Fixes #798

### DIFF
--- a/tune/change-csum.c
+++ b/tune/change-csum.c
@@ -79,10 +79,10 @@ static int check_csum_change_requreiment(struct btrfs_fs_info *fs_info, u16 new_
 		error("failed to check the dev-reaplce status: %m");
 		return ret;
 	}
-	if (ret == 0) {
-		error("running dev-replace detected, please finish or cancel it.");
-		return -EINVAL;
-	}
+//	if (ret == 0) {
+//		error("running dev-replace detected, please finish or cancel it.");
+//		return -EINVAL;
+//	}
 
 	if (fs_info->csum_type == new_csum_type) {
 		error("the fs is already using csum type %s (%u)",


tune/change-csum.c
```diff
@@ -79,10 +79,10 @@ static int check_csum_change_requreiment(struct btrfs_fs_info *fs_info, u16 new_
 		error("failed to check the dev-reaplce status: %m");
 		return ret;
 	}
-	if (ret == 0) {
-		error("running dev-replace detected, please finish or cancel it.");
```
Zygo, line 83: If ret is 0, we should check `replace_state` in the item and see if the state is acceptable:

* BTRFS_IOCTL_DEV_REPLACE_STATE_CANCELED
* BTRFS_IOCTL_DEV_REPLACE_STATE_FINISHED
* BTRFS_IOCTL_DEV_REPLACE_STATE_NEVER_STARTED (in this case I believe the dev_replace_item doesn't exist, but it's reasonable to accept this value anyway)

If the state is one of those, then there's no active replace in progress, and conversion can proceed.  Any other value for `replace_state` should error out (it's a bad value, or a new value and we don't know whether it's safe to proceed).

I would point to the tree printing code for an example of how to read a value out of an eb, but it turns out the tree printer doesn't print dev_replace items.

Would be nice to fix the spelling of "dev-reaplce" a few lines above...

adam900710:
Sorry, this is not a proper fix, it just completely ignore all dev-replace checks which is totally wrong.

Zygo's correct, we should do more checks instead.
Closing and would submit a proper fix instead.

Thanks for the report.

PR #807, opened 2024-06-04
Small bug fixes

adam900710:
This includes 3 bug fixes for btrfs-progs issues (all bugs), with one extra for misc/038 which is always causing false alerts here.


kdave:
Merged to devel, thanks.

PR #809, opened 2024-06-07
btrfs-progs: do not use async discard for misc/004

adam900710:
Finally I gave up skip misc/004, and spend some time on pinning down the root cause of the failure.


kdave:
Merged to devel, thanks.

PR #810, opened 2024-06-08
btrfs-progs: fix the conflicting super flags

adam900710:
There is a github issue that a canceled csum conversion leaves a
CHANGING_FSID_V2 flag, which turns out to be conflicting super flags
(CHANGING_FSID_V2 and CHANGING_DATA_CSUM).

Fix the problem and output all supported flags for print-tree.


kdave:
Merged to devel, thanks.

PR #811, opened 2024-06-10
Revert "btrfs-progs: convert: add raid-stripe-tree to allowed features"

adam900710:
This reverts commit 346a3819237b319985ad6042d6302f67b040a803.

The RST feature (at least for now) is mostly for zoned devices to support extra data profiles.

Thus btrfs-convert is never designed to handle RST, because there is no way an ext4/reiserfs can even be created on a zoned device, or it would create the correct RST tree root at all

Revert this unsupported feature to prevent false alerts.
---
This revert was not merged into v6.9, while it is included for the btrfs-convert unwritten extents patchset.


kdave:
I'll copy the reason from pull request #789:  I don't find the argument strong enough. RST is for zoned and will be the fix for RAID56 and can be also used independently (even if not bringing a huge improvement). The failing tests will need to handle it, we have multiple testing environments so this is a standard thing.

adam900710:
Again, the same problem, further more, just test it with `CONFIG_BTRFS_DEBUG` enabled 6.9 kernel and see how it fails.

I see no reason why you're so eager to push RST without any strong arugment.

PR #813, opened 2024-06-13
Hide rst behind experimental

adam900710:
Although we have exported raid-stripe-tree feature to end users,
the feature is still experimental as only kernels with
CONFIG_BTRFS_DEBUG can even mount it.

This results a feature mismatch in btrfs-progs and kernels, and can lead
to complains/confusion from end users.

This patchset would hide RST feature back behind experimental builds for
mkfs, and completely disable rst for convert, so end users won't and
can't enable RST by accident.

Now RST related test cases (mkfs/029, mkfs/030) are skipped for
regular builds, and a new test case (mkfs/033) is skipped for
experimental builds to make sure there is really no way to enable RST
support for non-experimental builds.

The new experimental build detection also provides the basis for
other experimental features like csum conversion in the future (recent
days).

Both regular and experimental builds have passed the full test suite
(except random misc/055 failure, which is caused by a clear-stale
behavior change, and will be addressed separately)


kdave:
Merged to devel, thanks.

PR #814, opened 2024-06-13
Misc/055 fix

adam900710:
[BUG]
There is a random chance that misc/055 would fail with stale qgroups
detected.

[CAUSE]
Commit 82f7d6c1d711 ("btrfs-progs: qgroup: handle stale qgroup deletion
more accurately") changed the behavior of "btrfs qgroup clear-stale"
that it will no longer try to remove qgroup undert deletion.

And the test case itself relies on the return value of "btrfs qgroup
clear-stale" to do the extra wait for subvolume deletion.

This means after that commit, the test case would skip the wait if there
is a subvolume waiting for cleanup, result a race window where the
subvolume can be dropped and become stale, and eventually trigger the
stale qgroup detection and cause false alerts.

[FIX]
Fix the test case by always wait for the subvolume to be dropped, so
that later "btrfs qgroup clear-stale" can properly remove all staled
qgroups.


kdave:
Merged to devel, thanks.

PR #815, opened 2024-06-13
Dump dmesg

adam900710:
Dump dmesg for mount failures. Mostly for the random CI misc/028 failure.


kdave:
Merged to devel, thanks.

PR #820, opened 2024-06-18
btrfs-progs: fix csum metadata reservation and add a basic test case for it

adam900710:
The current csum conversion is using an incorrect parameter for
btrfs_start_transaction(), which is 16K times larger than the expected
metadata space, thus it always fail with -ENOSPC.

Fix that first, then add a basic test case using the same populate_fs()
from convert test cases, and iterate through all the support csum
algorithms.


kdave:
Merged to devel, thanks.

PR #823, opened 2024-06-19
Data csum generation inject

adam900710:
@kdave  Please allow me to handle this case.

This one includes a fix that should be folded into a merged but yet released commit.

### DIFF
--- a/tune/change-csum.c
+++ b/tune/change-csum.c
@@ -371,29 +371,48 @@ static int generate_new_data_csums(struct btrfs_fs_info *fs_info, u16 new_csum_t
 	return generate_new_data_csums_range(fs_info, 0, new_csum_type);
 }
 
+/* After deleting/modifying this many leaves, commit a transaction. */
+#define CSUM_CHANGE_LEAVES_THRESHOLD	32
+
 static int delete_old_data_csums(struct btrfs_fs_info *fs_info)
 {
 	struct btrfs_root *csum_root = btrfs_csum_root(fs_info, 0);
-	struct btrfs_trans_handle *trans;
+	struct btrfs_trans_handle *trans = 0;
 	struct btrfs_path path = { 0 };
 	struct btrfs_key last_key;
+	unsigned int deleted_leaves = 0;
 	int ret;
 
 	last_key.objectid = BTRFS_EXTENT_CSUM_OBJECTID;
 	last_key.type = BTRFS_EXTENT_CSUM_KEY;
 	last_key.offset = (u64)-1;
 
-	trans = btrfs_start_transaction(csum_root, 1);
-	if (IS_ERR(trans)) {
-		ret = PTR_ERR(trans);
-		errno = -ret;
-		error("failed to start transaction to delete old data csums: %m");
-		return ret;
-	}
 	while (true) {
 		int start_slot;
 		int nr;
 
+		if (deleted_leaves >= CSUM_CHANGE_LEAVES_THRESHOLD) {
+			UASSERT(trans);
+			ret = btrfs_commit_transaction(trans, csum_root);
+			if (ret < 0) {
+				errno = -ret;
+				error_msg(ERROR_MSG_COMMIT_TRANS,
+					  "deleting old data csums: %m");
+				return ret;
+			}
+			trans = NULL;
+			deleted_leaves = 0;
+		}
+		if (!trans) {
+			trans = btrfs_start_transaction(csum_root, 1);
+			if (IS_ERR(trans)) {
+				ret = PTR_ERR(trans);
+				errno = -ret;
+				error_msg(ERROR_MSG_START_TRANS,
+					  "deleting old data csums: %m");
+				return ret;
+			}
+		}
 		ret = btrfs_search_slot(trans, csum_root, &last_key, &path, -1, 1);
 		if (ret < 0) {
 			errno = -ret;
@@ -428,6 +447,7 @@ static int delete_old_data_csums(struct btrfs_fs_info *fs_info)
 			break;
 		}
 		btrfs_release_path(&path);
+		deleted_leaves++;
 	}
 	btrfs_release_path(&path);
 	if (ret < 0)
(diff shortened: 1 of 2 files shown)


kdave:
Error injection works only when built with `-DINJECT`.

adam900710:
That's already taken into consideration.

The test would check if injection works for the very first commit transaction, if failed, the test case would be skipped.

kdave:
If the command fails for another reason then it cannot be distinguished if it's the injection or not, eg. when the injection would be enabled and it still fails.

This points out we need a way to determine if the injections are built in or not by querying the `btrfs` binary (--help or in `btrfs version`).

And then enable it in the CI workflow files (probably relevant for all). Right now it needs to be enabled by `EXTRA_CFLAGS=-DINJECT`, I'll do something more convenient.

kdave:
I like what eg. systemd does with the --version and feature description:

```
systemd 255 (255.6+suse.30.g3ea0e1dff3)
+PAM +AUDIT +SELINUX +APPARMOR +IMA -SMACK +SECCOMP +GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 +PWQUALITY +P11KIT +QRENCODE +TPM2 +BZIP2 +LZ4 +XZ +ZLIB +ZSTD +BPF_FRAMEWORK -XKBCOMMON -UTMP +SYSVINIT default-hierarchy=unified
```

We can do the same as there are several optional build features and it's not unified how this can be found out.

adam900710:
Personally speaking, I'd prefer to configure most things (including the *SAN and debug) into configure.

So that we can go the same `include/config.h` for everything.

adam900710:
For now, I will go with better stderr/stdout filter to make sure it's the injection got triggered and wait for a proper feature detection (either all configure or systemd like way).

adam900710:
And you only need to enable "D=1" for make, as INJECT is already included there.

kdave:
The D= could be done in configure too but as it is now it allows to do quick rebuilds of the same configured setup. It does not need to be tracked anywhere once done, then run default build, if it fails then rebuild with D=1 or anything else and check results.

kdave:
> For now, I will go with better stderr/stdout filter to make sure it's the injection got triggered

OK.

kdave:
The built-in feature list in `btrfs version` is now in devel.

kdave:
Please refresh the patches/PR, seems that it now conflicts.

adam900710:
Sure, will refresh the patchset and utilize the new btrfs built-in feature list instead.

tests/misc-tests/065-csum-conversion-inject/test.sh
```diff
@@ -0,0 +1,45 @@
...
+test_resume_data_csum_generation()
+{
+	local new_csum="$1"
+	local tmp=$(_mktemp "csum-convert")
+
+	# Error at the end of the data csum generation.
+	export INJECT="0x4de02239"
+	run_mustfail_stdout "error injection not working" \
+		"$TOP/btrfstune" --csum "$new_csum" "$TEST_DEV" &> $tmp
+	cat "$tmp" >> "$RESULTS"
```
kdave, line 21: I'd like to avoid any use of raw $RESULTS in the test cases, there should be helper for everything (like _log) but here we need a new one (_log_file), this is for a later cleanup.

tune/change-csum.c
```diff
@@ -371,29 +371,48 @@ static int generate_new_data_csums(struct btrfs_fs_info *fs_info, u16 new_csum_t
...
-		errno = -ret;
-		error("failed to start transaction to delete old data csums: %m");
-		return ret;
-	}
 	while (true) {
 		int start_slot;
 		int nr;
 
+		if (deleted_leaves >= CSUM_CHANGE_LEAVES_THRESHOLD) {
+			assert(trans);
```
kdave, line 395: Please don't use raw assert, there's UASSERT that honors the BTRFS_DISABLE_BACKTRACE and can also print stack trace.

tune/change-csum.c
```diff
@@ -445,27 +465,43 @@ static int delete_old_data_csums(struct btrfs_fs_info *fs_info)
...
-		errno = -ret;
-		error("failed to start transaction to change csum objectids: %m");
-		return ret;
-	}
 	while (true) {
 		struct btrfs_key found_key;
 		int nr;
 
+		if (changed_leaves >= CSUM_CHANGE_LEAVES_THRESHOLD) {
+			assert(trans);
```
kdave, line 484: same here, UASSERT

kdave (changes requested):

tests/misc-tests/065-csum-conversion-inject/test.sh
adam900710: I can do a cleanup later. As the pattern is utilized quite widely.

tune/change-csum.c
adam900710: Sure, and I guess the only reason we didn't just use `UASSERT()` to re-define `assert()` is, that `assert()` is a built-in function that can not be re-defined?

tests/misc-tests/065-csum-conversion-inject/test.sh
kdave: Fortunatelly it's not used widely, the other instance is in misc-tests/005-convert-progress-thread-crash/test.sh, the test infrastructure can use RESULTS freely.

kdave (approved):

tests/misc-tests/065-csum-conversion-inject/test.sh
adam900710: I mean not only the `_log_file`, but also some location that can be converted to `_log`.

E.g. all the `tests/*-test.sh` script are using something like `echo "=== START TEST $testname" >> "$RESULTS"`.
The same for a lot of internal usage inside `tests/common`.

PR #826, opened 2024-06-20
Mount detection fix

adam900710:
This one should be folded into the offending commit (not yet in any release).

I'll do a regular rebase and merge, but later fix it locally and force push it to devel.


adam900710:
Wrong base branch...

PR #827, opened 2024-06-20
btrfs-progs: tests: fix the mount failure detection

adam900710:
This fix should be folded into the offending one.

I'll do a rebase-and-merge first, then fold it locally, then a forced push.


adam900710:
The fix is folded into the offending commit now, and devel forced updated.

PR #828, opened 2024-06-21
btrfs-progs: do not check ram_bytes for non-compressed data extents

adam900710:
This patch reverts the following 3 commits:
d0cc40d23aa0 ("btrfs-progs: tests: add test case for ram_bytes detection and repair") 7313573c1942 ("btrfs-progs: check: original, detect and repair ram_bytes mismatch") 97bf7a596900 ("btrfs-progs: check: lowmem, detect and repair mismatched ram_bytes")

The problem with the ram_bytes check is, kernel can handle it without any problem, and the original objective for this is to detect such problem as I immaturelly believe the problem is fixed.

But it turns out to be incorrect and this check is already causing problems.

Fix it by doing a full revert for now.


kdave:
Merged to devel, thanks.

PR #832, opened 2024-07-01
btrfs-progs: scrub status: with --si, show rate in metric units

ivan:
This makes `btrfs scrub status --si` show the `Rate` in metric units as well.

before:

```
Total to scrub:   877.65GB
Rate:             609.22MiB/s
```

after:

```
Total to scrub:   877.65GB
Rate:             638.81MB/s
```


kdave:
Thanks, merged to devel.

PR #834, opened 2024-07-02
Cleanups

adam900710:
Thanks to Mark's new effort to introduce subvolume creation ability, the
long existing duplicated subvolume creation problem is exposed again.

The first patch to do a small cleanup for btrfs_create_tree() so that
the parameter list matches the kernel one.

The second one is the main dish to fully merge the different functions
to create a subvolume.

We have btrfs_create_tree() to properly create an empty tree, and
btrfs_make_root_dir() to create the initial root dir.

So use them to create btrfs_make_subvolume():

- Calls btrfs_create_tree() to properly create an empty tree
  Unlike btrfs_copy_root() used in create_subvol(), which can be unsafe
  if the source subvolume is not empty.

- Calls btrfs_read_fs_root() to setup the cache and tracking
  Inside create_data_reloc_tree() we directly added the root to
  fs_root_tree, which is only safe for data reloc tree.

  As we didn't properly set the correct tracking flags.

- Calls btrfs_make_root_dir() to setup the root directory
- Calls btrfs_update_root() to reflect the rootdir change

### DIFF
--- a/.github/workflows/devel.yml
+++ b/.github/workflows/devel.yml
@@ -28,7 +28,17 @@ jobs:
         run: ./autogen.sh && CC=${{ matrix.compiler }} ./configure
       - name: Documentation
         run: make V=1 -C Documentation
-      - name: Generate manual pages preview
+      - name: Generate manual pages preview (html)
+        if: ${{ matrix.compiler == 'gcc' }}
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
+        run: |
+          for file in ${ALL_CHANGED_FILES}; do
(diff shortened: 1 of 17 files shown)


adam900710:
Updated to version 2, with a new patch to clean up `btrfs_mksubvol()`.

kdave:
I've posted reveiw comments here but I don't see them fixed in the latest branch. It's possible that I'm reading it wrong in the github ui, it's comparing the changes to the actual devel branch and there are other changes too.

kdave (changes requested):
test

kdave (approved):
ok

adam900710:
Merged, close.

PR #839, opened 2024-07-10
DOC: remove stated restriction that swapfile must be on a single device FS

yarikoptic:
Seems to work fine as long as DATA profile is "single"

    # cat /proc/swaps
    Filename				Type		Size		Used		Priority
    /mnt/backup/swap.10G                    file		10485756	0		-2

    # btrfs fi usage /mnt/backup
    Overall:
        Device size:		 254.67TiB
        Device allocated:		 182.03TiB
        Device unallocated:		  72.64TiB
        Device missing:		     0.00B
        Device slack:		     0.00B
        Used:			 181.49TiB
        Free (estimated):		  72.81TiB	(min: 36.49TiB)
        Free (statfs, df):		  72.81TiB
        Data ratio:			      1.00
        Metadata ratio:		      2.00
        Global reserve:		 512.00MiB	(used: 0.00B)
        Multiple profiles:		        no

    Data,single: Size:178.52TiB, Used:178.35TiB (99.91%)
       /dev/mapper/vg_backup-lv_backup	 118.58TiB
       /dev/md17	  59.95TiB

    Metadata,RAID1: Size:1.75TiB, Used:1.57TiB (89.48%)
       /dev/mapper/vg_backup-lv_backup	   1.75TiB
       /dev/md17	   1.75TiB

    System,RAID1: Size:32.00MiB, Used:21.41MiB (66.89%)
       /dev/mapper/vg_backup-lv_backup	  32.00MiB
       /dev/md17	  32.00MiB

    Unallocated:
       /dev/mapper/vg_backup-lv_backup	  61.58TiB
       /dev/md17	  11.06TiB

### DIFF
--- a/.github/workflows/devel.yml
+++ b/.github/workflows/devel.yml
@@ -28,7 +28,17 @@ jobs:
         run: ./autogen.sh && CC=${{ matrix.compiler }} ./configure
       - name: Documentation
         run: make V=1 -C Documentation
-      - name: Generate manual pages preview
+      - name: Generate manual pages preview (html)
+        if: ${{ matrix.compiler == 'gcc' }}
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
+        run: |
+          for file in ${ALL_CHANGED_FILES}; do
+            echo "$file was changed, generate preview to summary"
+            Documentation/html-preview.sh "$file" >> $GITHUB_STEP_SUMMARY
+          done
+      - run: echo '<hr>' >> $GITHUB_STEP_SUMMARY
+      - name: Generate manual pages preview (man)
         if: ${{ matrix.compiler == 'gcc' }}
         env:
           ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
--- a/Documentation/btrfs-inspect-internal.rst
+++ b/Documentation/btrfs-inspect-internal.rst
@@ -228,6 +228,9 @@ tree-stats [options] <device>
         -b
                 Print raw numbers in bytes.
 
+        -t <treeid>
+                Print stats only for the given treeid.
+
 EXIT STATUS
 -----------
 
--- a/Documentation/ch-swapfile.rst
+++ b/Documentation/ch-swapfile.rst
@@ -6,7 +6,8 @@ restrictions for active swapfiles don't apply.
 There are some limitations of the implementation in BTRFS and Linux swap
 subsystem:
 
-* filesystem - must be only single device
+* filesystem - should be only single device (depending on setup might not work
+  on multiple)
 * filesystem - must have only *single* data profile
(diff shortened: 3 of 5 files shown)


kdave:
This is probably a conincidence and it is possible to create the swap file in that way but it's not generally reliable. As long as the swapfile extents are on one device (and single profile, multi-device filesystem), it works. https://elixir.bootlin.com/linux/latest/source/fs/btrfs/inode.c#L10853 : 

```
10698                 } else if (device != map->stripes[0].dev) {                                                                                                                                                    
10699                         btrfs_warn(fs_info, "swapfile must be on one device");                                                                                                                                 
10700                         ret = -EINVAL;                                                                                                                                                                         
10701                         goto out;                                                                                                                                                                              
10702                 } 
```

How exactly is the swapfile laid out on the devices cannot be directly affected from user space. On a fragmented free space it's likely that extents end up on different devices, on the chunk (1G) level it depends on the free space on each device where allocator picks the the next one. The chances to fit a swap file on one device also decreases with its size.

The documentation could be more specific about that, in the sense "try it but it may not work", but removing the constraint will lead to somebody filing a report that it does not work.

yarikoptic:
Thanks for analysis! I will close PR for now .

kdave:
Ok, no problem. You could keep it open even if the fix is incomplete or needs updates, if you found an issue to fix the PR is a reminder and it can be fixed partially or there's more to uncover. What I wrote in the comment could be put into the documentation, formatting that and adapting to the style documentation uses would be a good contribution.

yarikoptic:
Ok, will re-approach when get to keyboard...

yarikoptic:
rewritten, changing `must` to `should` and adding brief clarification

manueldeprada:
@yarikoptic are you able to use that swapfile for hibernation with multidevice filesystem? My swapfile also mounts fine, but I am unable to hibernate.

I tried to debug but it's not clear to me what to write in /sys/power/resume

yarikoptic:
Sorry, don't know :-(  I wouldn't dare to hibernate that monster.

kdave:
> I tried to debug but it's not clear to me what to write in /sys/power/resume

It's `/sys/power/resume_offset`, there should be a physical offset of the swapfile, that can be obtained by `btrfs inspect map-swapfile`.

manueldeprada:
> > I tried to debug but it's not clear to me what to write in /sys/power/resume
> 
> It's `/sys/power/resume_offset`, there should be a physical offset of the swapfile, that can be obtained by `btrfs inspect map-swapfile`.

I tried that. If I do it, `echo disk > /sys/power state` fails with a "no such device". systemctl hibernate fails with `Failed to find location to hibernate to: Stale file handle` or with `Call to Hibernate failed: Invalid argument`. 

Then I found that `/sys/power/resume` is set to `0:0`. That looked very wrong and that's why I thought it could be the problem.

kdave:
I wrote a paragraph summing up this discussion.

PR #844, opened 2024-07-15
btrfs-progs: convert: fix the rollback filename output

adam900710:
When testing my newer version of subvolume creation cleanup (which
removes btrfs_mksubvol(), utilized by convert), I found out that
"btrfs-convert -r" output is always corrupted.

It turns out to be a small string termination problem.

The first patch fixes it, then a new test case for it.

### DIFF
--- a/tests/convert-tests/026-rollback-output/test.sh
+++ b/tests/convert-tests/026-rollback-output/test.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+# Make sure "btrfs-convert -r" is outputting the correct filename
+
+source "$TEST_TOP/common" || exit
+source "$TEST_TOP/common.convert" || exit
+
+setup_root_helper
+prepare_test_dev
+
+check_global_prereq mkfs.ext4
+check_prereq btrfs-convert
+check_prereq btrfs
+
+convert_test_prep_fs ext4 mke2fs -t ext4 -b 4096
+run_check_umount_test_dev
+convert_test_do_convert
+
+# Rollback and save the output.
+run_check_stdout "$TOP/btrfs-convert" --rollback "$TEST_DEV" |\
+	grep -q "ext2_saved/image" || _fail "rollback filename output is corruptedd"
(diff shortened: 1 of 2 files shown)


kdave (changes requested):

tests/convert-tests/026-rollback-output/test.sh
```diff
@@ -0,0 +1,26 @@
...
+
+check_global_prereq mkfs.ext4
+check_prereq btrfs-convert
+check_prereq btrfs
+
+convert_test_prep_fs ext4 mke2fs -t ext4 -b 4096
+run_check_umount_test_dev
+convert_test_do_convert
+
+tmp=$(mktemp --tmpdir btrfs-progs-convert-rollback.XXXXXX)
```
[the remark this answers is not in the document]
adam900710: Buy here we only want a small tmpfile, not a directory. I'll go `_mktemp()` instead.

tests/convert-tests/026-rollback-output/test.sh
```diff
@@ -0,0 +1,26 @@
...
+
+convert_test_prep_fs ext4 mke2fs -t ext4 -b 4096
+run_check_umount_test_dev
+convert_test_do_convert
+
+tmp=$(_mktemp convert-rollback)
+# Rollback and save the output.
+run_check_stdout "$TOP/btrfs-convert" --rollback "$TEST_DEV" >> "$tmp"
+
+if ! grep -q "ext2_saved/image" "$tmp"; then
```
kdave, line 22: Using a temporary file is ok, it makes the code a bit more readable, otherwise it can be also in a pipe like is eg. in tests/cli-tests/002-balance-full-no-filters/test.sh:
```
 20 run_check_stdout $SUDO_HELPER "$TOP/btrfs" balance start --background "$TEST_MNT" |
 21         grep -F "Full balance without filters requested." >/dev/null ||
 22         _fail "full balance warning not in the output"
```

kdave (approved):
One comment regarding style but otherwise ok.

tests/convert-tests/026-rollback-output/test.sh
adam900710: Normally I'd prefer avoid long pipes (IIRC some bash setting can change the return value for long pipes), but in this particular case, it's only two commands and I'm fine to put them into the `if` condition.

Will fix it in the next (and hopefully the last) update.

PR #858, opened 2024-07-31
btrfs-progs: rework how we traverse rootdir

adam900710:
Thanks to Mark's recent work, I finally get some time to rework rootdir
traversal.

All the problems are described inside the second patch.
While the last patch is a small enhancement to --rootdir to reject hard
links.

With this change, it's much easier to support subvolume creations at
mkfs time:

- Create a hashmap (or other similar structure) to record all the
  directories that should be subvolume

- Call btrfs_make_subvoume() other than btrfs_insert_inode() if a path
  should be a subvolume

- Call btrfs_link_subvolume() other than btrfs_add_link() for a
  subvolume

Everything like parent directory inode size is properly handled by
btrfs_link_subvolume() and btrfs_add_link() already.

### DIFF
--- a/kernel-shared/ctree.h
+++ b/kernel-shared/ctree.h
@@ -1210,7 +1210,7 @@ int btrfs_new_inode(struct btrfs_trans_handle *trans, struct btrfs_root *root,
 int btrfs_change_inode_flags(struct btrfs_trans_handle *trans,
 			     struct btrfs_root *root, u64 ino, u64 flags);
 int btrfs_add_link(struct btrfs_trans_handle *trans, struct btrfs_root *root,
-		   u64 ino, u64 parent_ino, char *name, int namelen,
+		   u64 ino, u64 parent_ino, const char *name, int namelen,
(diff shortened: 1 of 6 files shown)


kdave (changes requested):

adam900710:
Extra changelog, as some feedback from Boris is also addressed.

[CHANGELOG]
v2:
- Change the current_path.level, so that 0 means uninitialized
  This makes the current_path level to match the ftwbuf level.

- Add a comment explaining how the current_path stack works
  With an example layout and stack changes.

- Add two new test cases
  Both test cases are for the old rootdir bugs:
  * Extra hard link are out of the rootdir
    Old --rootdir will create a corrupted fs with incorrect nlink.
    This is because we use st_nlink without any extra verification

  * Conflicting inode numbers caused by different mount points
    Old --rootdir will fail gracefully, but still not ideal as the error
    message doesn't explain it at all.
    This is because we use st_inode without any extra verification

- Newline and typo fixes

kdave (approved):

PR #861, opened 2024-08-02
btrfs-progs: add recursive subvol delete

adam900710:
Adds a --recursive option to btrfs subvol delete, causing it to pass the BTRFS_UTIL_DELETE_SUBVOLUME_RECURSIVE flag through to libbtrfsutil.

This is a resubmission of part of a patch that Omar Sandoval sent in 2018, which appears to have been overlooked:
https://lore.kernel.org/linux-btrfs/e42cdc5d5287269faf4d09e8c9786d0b3adeb658.1516991902.git.osandov@fb.com/

Reviewed-by: Qu Wenruo <wqu@suse.com>

### DIFF
--- a/Documentation/btrfs-subvolume.rst
+++ b/Documentation/btrfs-subvolume.rst
@@ -112,6 +112,16 @@ delete [options] [<subvolume> [<subvolume>...]], delete -i|--subvolid <subvolid>
         -i|--subvolid <subvolid>
                 subvolume id to be removed instead of the <path> that should point to the
                 filesystem with the subvolume
+
+        -R|--recursive
+		delete subvolumes beneath each subvolume recursively.
+
+		This requires either `CAP_SYS_ADMIN` or the fs is mounted with
+		`user_subvol_rm_allowed` mount option.
+		In the unprivileged case, subvolumes which cannot be accessed
+		are skipped.
+		The deletion is not atomic.
+
         -v|--verbose
                 (deprecated) alias for global *-v* option
 
--- a/cmds/subvolume.c
+++ b/cmds/subvolume.c
@@ -347,6 +347,7 @@ static const char * const cmd_subvolume_delete_usage[] = {
 	OPTLINE("-c|--commit-after", "wait for transaction commit at the end of the operation"),
 	OPTLINE("-C|--commit-each", "wait for transaction commit after deleting each subvolume"),
 	OPTLINE("-i|--subvolid", "subvolume id of the to be removed subvolume"),
+	OPTLINE("-R|--recursive", "delete subvolumes beneath each subvolume recursively"),
 	OPTLINE("-v|--verbose", "deprecated, alias for global -v option"),
 	HELPINFO_INSERT_GLOBALS,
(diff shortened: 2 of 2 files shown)


Documentation/btrfs-subvolume.rst
```diff
@@ -112,6 +112,10 @@ delete [options] [<subvolume> [<subvolume>...]], delete -i|--subvolid <subvolid>
         -i|--subvolid <subvolid>
                 subvolume id to be removed instead of the <path> that should point to the
                 filesystem with the subvolume
+
+        -R|--recursive
+                delete subvolumes beneath each subvolume recursively
+
```
kdave, line 118: This should be very clear that it either requires CAP_SYS_ADMIN to delete all subvolumes or mount option user_subvol_rm_allowed or that it will not delete inaccessible subvoluems. Basically what doucmentation btrfs_util_delete_subvolume says in libbtrfsutil/README.md. Same for help text.

cmds/subvolume.c
```diff
@@ -347,6 +347,7 @@ static const char * const cmd_subvolume_delete_usage[] = {
 	OPTLINE("-c|--commit-after", "wait for transaction commit at the end of the operation"),
 	OPTLINE("-C|--commit-each", "wait for transaction commit after deleting each subvolume"),
 	OPTLINE("-i|--subvolid", "subvolume id of the to be removed subvolume"),
+	OPTLINE("-R|--recursive", "delete subvolumes beneath each subvolume recursively"),
```
kdave, line 350: This must mention the conditions or limitations when it will not remove the subvolumes.

kdave (changes requested):

cmds/subvolume.c
adam900710: I do not think it's a good idea to put all details into the normally-one-line help line.

The details including the privilege/mount option requirement should go the man page.
Or every one-line help string will be no different than the man page.

kdave: Missing key information about the behaviour of the option is serious problem in usability, nothing even hints that there are exceptions so no reason to look to the manual page. The text does not need to be on one line, the OPTLINE is there to format the text and there are options with long text, eg 'check --force'.

adam900710: Your argument doesn't stand, especially there is *no* mention in the `btrfs subvolume delete` help string at all, on the privilege nor the mount option.

In fact the privilege/mount option requirement are all the same as regular subvolume deletion.
You're asking for details that should be included as generic requirement to be added to a specific option.

And even your privilege/mount option argument makes no sense for the recursive option.
Just go check the help string from `rm`. Does it even mention anything like privilege?

Let me be clear, if you want to add mount option/privilege for subvolume deletion, go add it to the generic help string of 'btrfs subvolume delete', not to this specific option.

If you want the recursive specific implementation details, I'm fine with the non-atomic part into the man page.

If you want the privilege/mount option/non-atomic details to be all added into help string, I'll call it BS.

kdave (approved):

Issue #606, opened 2023-03-22
btrfs receive occasionally fails to find parent subvolume when data flows in the reverse direction


ettavolt:
I'm trying to carry user profile between two laptops via _backups_ on an external disk (so that I have sync and backups and the same time).

For that I'm trying to utilize the following structure:

profile, btrfs subvolume:
 `/home/ask-rbr`
btrfs ro snapshots:
 `/home/.ask-rbr-snapshots/@yyyy-MM-dd[Thh[:mm[:ss]]]`
external drive, btrfs ro snapshots:
 `/run/rbrssd/ask-arch-profile/@yyyy-MM-dd[Thh[:mm[:ss]]]`

Both systems are running Arch with:
 linux 6.2.7.arch1-1
 btrfs-progs 6.2.1-1
(one might be lagging by a week).

This is an extension of earlier backup-only setup, when `send | receive` was going profile → external drive direction only.

The problem happens when I try to send recent backup to laptop to recreate profile (making it a rw snapshot of most recent ro snapshot).

External drive:
```
ask-arch-profile/@2023-03-16T11
 Name: @2023-03-16T11
 UUID: 88e0644b-6542-384f-9498-980582b29226
 Parent UUID: -
 Received UUID: cab9c40b-98dd-ca49-98d4-6271e8cf4fac
 Creation time: 2023-03-21 17:42:34 -0400
 Subvolume ID: 291
 Generation: 212
 Gen at creation: 209
 Parent ID: 5
 Top level ID: 5
 Flags: readonly
 Send transid: 42
 Send time: 2023-03-21 17:42:34 -0400
 Receive transid: 210
 Receive time: 2023-03-21 17:43:05 -0400
 Snapshot(s):
 ask-arch-profile/@2023-03-21

ask-arch-profile/@2023-03-21
	Name: 			@2023-03-21
	UUID: 			918fa20d-d6a5-724e-8f09-5b4aeeb22844
	Parent UUID: 		88e0644b-6542-384f-9498-980582b29226
	Received UUID: 		4bb66e19-4e51-9e49-ac0f-1574419be949
	Creation time: 		2023-03-21 17:43:24 -0400
	Subvolume ID: 		292
	Generation: 		215
	Gen at creation: 	212
	Parent ID: 		5
	Top level ID: 		5
	Flags: 			readonly
	Send transid: 		1553
	Send time: 		2023-03-21 17:43:24 -0400
	Receive transid: 	213
[...]

PR #643, opened 2023-06-18
btrfs-progs: receive: cannot find clone source subvol when receiving in reverse direction

ettavolt:
process_clone, unlike process_snapshot, only searched a subvolume matching by received uuid. However, when earlier "receiver" side sends, it mentions received uuid, which is for earlier "send" side (now "receiver" side) is just uuid.

Fixes #606.

PR #862, opened 2024-08-05
btrfs-progs: receive: cannot find clone source subvol when receiving in reverse direction

ettavolt:
process_clone() only searches the received_uuid, but could exist in an earlier uuid that isn't the received_uuid.  Mirror what process_snapshot does and search both the received_uuid and if that fails look up by normal uuid.

Fixes: https://github.com/kdave/btrfs-progs/issues/606

Issue: #606
Pull-request: #643

### DIFF
--- a/cmds/receive.c
+++ b/cmds/receive.c
@@ -221,6 +221,25 @@ out:
 	return ret;
 }
 
+/*
+ * Search for the @subvol_uuid, try received_uuid and subvolume as a fallback
+ * if it's not found.
+ */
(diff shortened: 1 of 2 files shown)


kdave:
Thanks, merged to devel. I have checked why this was not merged the first time, seems that I messed up rebasing my local branches and lost this and a few more patches. There are 2 changes under commit 1ebdb7dc178060d4cf5cfaa81b957d936eac8407.

Issue #816, opened 2024-06-13
checksum verify failure ignored as "no error found" 


tribbloid:
Encountered on v6.2 (Debian 12) and v6.7 (Fedora 40)

sample output:

```
$ sudo btrfs check --repair --check-data-csum /dev/disk/by-label/Home
enabling repair mode
WARNING:

	Do not use --repair unless you are advised to do so by a developer
	or an experienced user, and then only after having accepted that no
	fsck can successfully repair all types of filesystem corruption. E.g.
	some software or hardware bugs can fatally damage a volume.
	The operation will start in 10 seconds.
	Use Ctrl-C to stop it.
10 9 8 7 6 5 4 3 2 1
Starting repair.
Opening filesystem to check...
Checking filesystem on /dev/disk/by-label/Home
UUID: a5c81116-78a5-4edb-b57c-b08e90e1391b
[1/7] checking root items
Fixed 0 roots.
[2/7] checking extents
checksum verify failed on 58720256 wanted 0x8f087114 found 0xb87d3f03
No device size related problem found
[3/7] checking free space cache
[4/7] checking fs roots
[5/7] checking csums against data
[6/7] checking root refs
[7/7] checking quota groups skipped (not enabled on this FS)
found 423218954240 bytes used, no error found
total csum bytes: 19136412
total tree bytes: 2665676800
total fs tree bytes: 2427846656
total extent tree bytes: 210829312
btree space waste bytes: 549918529
file data blocks allocated: 488460259328
 referenced 428168486912
```

"no error found" is self-contradicting with "checksum verify failed on 58720256 wanted 0x8f087114 found 0xb87d3f03"

also no repair was attempted despite --repair option is used

adam900710:
No csum repair support, so it will do nothing.

And I tried it with latest v6.9 progs, it reports it correctly as an error, no matter if it's --repair:

```
[adam@btrfs-vm ~]$ btrfs check --check-data-csum --repair --force /dev/test/scratch1
enabling repair mode
Opening filesystem to check...
Checking filesystem on /dev/test/scratch1
UUID: 64e210b4-34f1-4b47-98cf-52ce991841e2
[1/7] checking root items
Fixed 0 roots.
[2/7] checking extents
No device size related problem found
[3/7] checking free space tree
[4/7] checking fs roots
[5/7] checking csums against data
mirror 1 bytenr 298848256 csum 0x13fec125 expected csum 0x98757625
ERROR: errors found in csum tree
[...]

Issue #482, opened 2022-06-18
Enhancement: scrub on nodatasum files


Forza-tng:
An improvement to scrub would be to do a bytewise compare of nodatasum files if `DUP`, `RAID1/10/c34` is used. This can help detect possible corruption if one of the copies have mismatch the other.

An enhanced `btrfs restore` tool could then be used to extract all copies so that the user could compare the files herself.

In the case of RAID1/10c34, scrub could also help identify which copies that are likely the bad/good ones.

RAID56. Perhaps it would be possible to check the data against the parity stripe to detect corruption, even though it can't say if the parity of the data is correct.

Issue #134, opened 2018-05-13
Feature Requst: mini-scrub mirrored NoCOW data


jamespharvey20:
On a 3 device btrfs RAID 1, I have 2 files that are marked NoCOW, but are still lzo compressed, that when read crash the system.  They're journald files, so nothing that should crash the system.  This has been discussed on the mailing list, with title: '"decompress failed" in 1-2 files always causes kernel oops, check/scrub pass'.  Chris Murphy remembered in the archives a discussion that compression can be forced on NoCOW under certain circumstances.  Bug reported here: https://bugzilla.kernel.org/show_bug.cgi?id=199707

Being marked NoCOW, they don't have checksums, so btfs scrub doesn't look at them.

In my situation, one of the mirrored copies is valid, and the other is invalid.

Even once the kernel crash is fixed, it would be a really important data integrity feature for btrfs-progs (I'm thinking a new feature on scrub) that when running on a mirrored volume to look at any files without a checksum, and compare the mirrored copies.  If they're different, obviously nothing can be automatically corrected without a checksum to verify.  But, it's important to let the user know there's a problem.  They can check which version appears valid, restore that file from backups, just know something is wrong, etc.

adam900710:
The problem is, how such compressed file is generated.
Normally for NODATACOW file we shouldn't generate compressed file extent.

So I'd say we need to know how such misbehavior happens and then fix kernel for it (both btrfs and lzo).
And for already genreated compressed extent without csum, enhance btrfs check to expose them.

adam900710:
Pull request https://github.com/kdave/btrfs-progs/pull/135 should address the problem, although not the direction of fixing/scrubbing them.

That pull request will report such extents as errors, and for kernel btrfs fix, at least we will prevent such problems by never compress any extent if NODATASUM is set.

jamespharvey20:
I agree with #135.

Wanted to clarify if you think #135 makes the feature request unnecessary.

I still think there's good reason for it.  Although the kernel patches and #135 will prevent NODATASUM/NODATACOW data from being compressed, there can of course still be uncompressed data without checksums that has mirrored copies that could have one copy get corrupted.  If the good copy is read, user never gets alerted their data is no longer actually mirrored, unless the good copy gets corrupted too.  Then (or, if the bad copy is read initially before the good copy) user silently gets bad data.

If the mirrored copies are compared, user will know something's wrong and be able to deal with it, granted not automatically, when it's still fixable.

[...]

PR #863, opened 2024-08-06
Improve documentation around userspace tool behaviour and safety

csnover:
Hi there,

On the [mailing list](https://lore.kernel.org/linux-btrfs/02afb6b7-e566-474e-80ff-3ace99693553@zetafleet.com/T/) for the past few days I have been shouting into the void about some issues I encountered trying to deal with a hardware failure on a btrfs raid-1 filesystem. In my last mail I made several recommendations, and while I won’t be able to become familiar enough with the internals of the btrfs tools themselves to send patches for some of them, but I am at least capable of writing a few paragraphs of documentation. So, here are some patches for that.

In the scrub documentation change I made, I am not sure if there are other edge cases beyond the split-brain one I document where running scrub would be dangerous. I adapted the line about its safety from the comment at https://github.com/kdave/btrfs-progs/issues/816#issuecomment-2166698048. I think that comment probably contains more useful information that could be integrated into the documentation too, but I am currently out of energy to make additional changes.

I also think I may be documenting some bugs here, like that btrfs-check looks at metadata from some place other than the specified block device which seems like it should not be doing since it does not take a filesystem path, but at least I would prefer the current behaviour to be documented than to have one other person be shocked when they see errors that go away after a read-write scrub.

I hope these changes are helpful. Please feel free to change or adapt as necessary, I do not anticipate focusing much more on this right now because I have other obligations.

Thank you for writing and maintaining btrfs and I hope you have a great week!

### DIFF
--- a/Documentation/btrfs-check.rst
+++ b/Documentation/btrfs-check.rst
@@ -19,6 +19,12 @@ by the option *--readonly*.
 
 :command:`btrfsck` is an alias of :command:`btrfs check` command and is now deprecated.
 
+.. note::
+   Even though the filesystem checker requires a device argument, it scans for all
+   devices belonging to the same filesystem and may report metadata errors from other
+   devices that are correctable by :command:`btrfs scrub`. In this case, run scrub
+   first to ensure any correctable metadata errors are fixed to avoid false-positives.
+
 .. warning::
    Do not use *--repair* unless you are advised to do so by a developer
    or an experienced user, and then only after having accepted that no *fsck*
--- a/Documentation/ch-scrub-intro.rst
+++ b/Documentation/ch-scrub-intro.rst
@@ -7,13 +7,26 @@ also automatically repair any damage by default by copying verified good data
 from one of the other replicas.
 
 .. warning::
+   Files with the NOCOW (``chattr +C``) attribute imply ``nodatasum``.
+   Errors in these files cannot be detected or corrected by scrub. This means
+   the NOCOW attribute is not currently safe in the presence of replicas since
+   btrfs cannot determine which copy of data to use after a failure and may
+   return data from any replica, including corrupt or outdated data.
+   Detecting and recovering from a failure in this case requires manual
+   intervention. See issue `#482 <https://github.com/kdave/btrfs-progs/issues/482>`_.
+   
+   Notably, `systemd sets NOCOW on journals by default <https://github.com/systemd/systemd/commit/11689d2a021d95a8447d938180e0962cd9439763>`_,
+   and `libvirt ≥ 6.6 sets NOCOW on storage pool directories by default <https://www.libvirt.org/news.html#v6-6-0-2020-08-02>`_.
+   Other distributions may also enable NOCOW on database files or directories to
+   try to improve performance.
+
+.. warning::
    A read-write scrub will do no further harm to a damaged filesystem if it is not
    possible to perform a correct repair, so it is safe to use at almost any time.
-
-   The only time a read-write scrub is unsafe is during a split-brain event. This
-   is unlikely and requires a specific sequence of events that cause an
-   unhealthy device or device set to be mounted read-write in the absence of the
-   healthy device or device set from the same filesystem. For example:
+   However, if a split-brain event occurs, btrfs scrub may cause unrecoverable data
+   loss. This situation is unlikely and requires a specific sequence of events that
+   cause an unhealthy device or device set to be mounted read-write in the absence
+   of the healthy device or device set from the same filesystem. For example:
 
    1. Device set F fails and drops from the bus, while device set H continues to
       function and receive additional writes.
@@ -33,8 +46,8 @@ from one of the other replicas.
    :doc:`btrfs-check` performs more exhaustive checking and can sometimes be
    used, with expert guidance, to rebuild certain corrupted filesystem structures
    in the absence of any good replica. However, when a replica exists, scrub is
-   able to automatically correct most errors reported by `btrfs-check`, so should
-   normally be run first to avoid false positives from `btrfs-check`.
+   able to automatically correct most errors reported by ``btrfs-check``, so should
+   normally be run first to avoid false positives from ``btrfs-check``.
 
 The user is supposed to run it manually or via a periodic system service. The
 recommended period is a month but it could be less. The estimated device bandwidth
--- a/Documentation/ch-scrub-intro.rst
+++ b/Documentation/ch-scrub-intro.rst
@@ -7,18 +7,30 @@ also automatically repair any damage by default by copying verified good data
 from one of the other replicas.
 
 .. warning::
-   Files with the NOCOW (``chattr +C``) attribute imply ``nodatasum``.
-   Errors in these files cannot be detected or corrected by scrub. This means
-   the NOCOW attribute is not currently safe in the presence of replicas since
-   btrfs cannot determine which copy of data to use after a failure and may
-   return data from any replica, including corrupt or outdated data.
-   Detecting and recovering from a failure in this case requires manual
-   intervention. See issue `#482 <https://github.com/kdave/btrfs-progs/issues/482>`_.
-   
-   Notably, `systemd sets NOCOW on journals by default <https://github.com/systemd/systemd/commit/11689d2a021d95a8447d938180e0962cd9439763>`_,
-   and `libvirt ≥ 6.6 sets NOCOW on storage pool directories by default <https://www.libvirt.org/news.html#v6-6-0-2020-08-02>`_.
-   Other distributions may also enable NOCOW on database files or directories to
-   try to improve performance.
+   Setting the ``No_COW`` (``chattr +C``) attribute on a file implicitly enables
+   ``nodatasum``. This means that while metadata for these files continues to
+   be validated and corrected by scrub, the actual file data is not.
+
+   Furthermore, btrfs does not currently mark missing or failed disks as
+   unreliable, so will continue to load-balance reads to potentially damaged
+   replicas in a replicated filesystem. This is not a problem normally because
+   damage is detected by checksum validation and a mirror copy is used, but
+   because ``No_COW`` files are not protected by checksum, bad data may be
+   returned even if a good copy exists on another replica. Which replica is used
+   is based on the process ID of the executable reading the file.
+
+   Writing to a ``No_COW`` file after reading from a bad replica will overwrite
+   all replicas with the bad data. Detecting and recovering from a failure in
+   this case requires manual intervention before the file is rewritten to avoid
+   data loss. See issue `#482 <https://github.com/kdave/btrfs-progs/issues/482>`_.
+   Even with raid1c3 or higher, for performance reasons, btrfs does not use
+   consensus reads on any files, even ``No_COW`` files, to validate or correct
+   data errors.
+
+   Notably, `systemd sets +C on journals by default <https://github.com/systemd/systemd/commit/11689d2a021d95a8447d938180e0962cd9439763>`_,
+   and `libvirt ≥ 6.6 sets +C on storage pool directories by default <https://www.libvirt.org/news.html#v6-6-0-2020-08-02>`_.
+   Other applications or distributions may also set +C to try to improve
+   performance.
 
 .. warning::
    A read-write scrub will do no further harm to a damaged filesystem if it is not
--- a/Documentation/ch-scrub-intro.rst
+++ b/Documentation/ch-scrub-intro.rst
@@ -17,7 +17,9 @@ from one of the other replicas.
    damage is detected by checksum validation and a mirror copy is used, but
    because ``No_COW`` files are not protected by checksum, bad data may be
    returned even if a good copy exists on another replica. Which replica is used
-   is based on the process ID of the executable reading the file.
+   is determined by the setting in ``/sys/fs/btrfs/<uuid>/read_policy``.
+   Currently, the only possible value for this setting is ``pid``, which uses
+   the process ID of the executable reading the file to pick the replica.
 
    Writing to a ``No_COW`` file after reading from a bad replica will overwrite
    all replicas with the bad data. Detecting and recovering from a failure in
--- a/Documentation/ch-scrub-intro.rst
+++ b/Documentation/ch-scrub-intro.rst
@@ -1,14 +1,40 @@
-Scrub is a pass over all filesystem data and metadata and verifying the
-checksums. If a valid copy is available (replicated block group profiles) then
-the damaged one is repaired. All copies of the replicated profiles are validated.
+Scrub is a validation pass over all filesystem data and metadata that detects
+checksum errors, super block errors, metadata block header errors, and disk
+read errors. All copies of replicated profiles are validated by default.
+
+On filesystems that use replicated block group profiles (e.g. raid1), scrub will
+also automatically repair any damage by default by copying verified good data
+from one of the other replicas.
+
+.. warning::
+   A read-write scrub will do no further harm to a damaged filesystem if it is not
+   possible to perform a correct repair, so it is safe to use at almost any time.
+
+   The only time a read-write scrub is unsafe is during a split-brain event. This
+   is unlikely and requires a specific sequence of events that cause an
+   unhealthy device or device set to be mounted read-write in the absence of the
+   healthy device or device set from the same filesystem. For example:
+
+   1. Device set F fails and drops from the bus, while device set H continues to
+      function and receive additional writes.
+   2. After a reboot, healthy set H does not reappear immediately, but failed set
+      F does.
+   3. Failed set F is mounted read-write. At this point, it is no longer safe for
+      set H to reappear as the transaction histories have diverged. Allowing set H
+      and set F to recombine at any point will cause corruption of set H. Running
+      scrub on a split-brained filesystem will overwrite good data from set H with
+      other data from set F, increasing the amount of permanent data loss.
 
 .. note::
-   Scrub is not a filesystem checker (fsck) and does not verify nor repair
-   structural damage in the filesystem. It really only checks checksums of data
-   and tree blocks, it doesn't ensure the content of tree blocks is valid and
-   consistent. There's some validation performed when metadata blocks are read
-   from disk (:doc:`Tree-checker`) but it's not extensive and cannot substitute
-   full :doc:`btrfs-check` run.
+   Scrub is not a filesystem checker (fsck). It can only detect filesystem damage
+   using the (:doc:`Tree-checker`) and checksum validation, and it can only repair
+   filesystem damage by copying from other known good replicas.
+
+   :doc:`btrfs-check` performs more exhaustive checking and can sometimes be
+   used, with expert guidance, to rebuild certain corrupted filesystem structures
+   in the absence of any good replica. However, when a replica exists, scrub is
+   able to automatically correct most errors reported by `btrfs-check`, so should
+   normally be run first to avoid false positives from `btrfs-check`.
 
 The user is supposed to run it manually or via a periodic system service. The
 recommended period is a month but it could be less. The estimated device bandwidth
--- a/Documentation/trouble-index.rst
+++ b/Documentation/trouble-index.rst
@@ -9,8 +9,9 @@ for description and may need further explanation what needs to be done.
 Error: parent transid verify error
 ----------------------------------
 
-Reason: result of a failed internal consistency check of the filesystem's metadata.
-Type: permanent
+| Reason: result of a failed internal consistency check of the filesystem's metadata.
+| Type: correctable by ``btrfs-scrub`` if a good copy exists on another replica; otherwise, permanent
+|
 
 .. code-block:: none
 
@@ -21,17 +22,20 @@ contains target block offset and generation that last changed this block. The
 block it points to then upon read verifies that the block address and the
 generation matches. This check is done on all tree levels.
 
-The number in **faled on 30736384** is the logical block number, **wanted 10**
+The number in **failed on 30736384** is the logical block number, **wanted 10**
 is the expected generation number in the parent node, **found 8** is the one
 found in the target block.  The number difference between the generation can
 give a hint when the problem could have happened, in terms of transaction
 commits.
 
-Once the mismatched generations are stored on the device, it's permanent and
-cannot be easily recovered, because of information loss. The recovery tool
-``btrfs restore`` is able to ignore the errors and attempt to restore the data
-but due to the inconsistency in the metadata the data need to be verified by the
-user.
+Once the mismatched generations are stored on the device, without a good copy
+from another replica, it's permanent and cannot be easily recovered because of
+information loss. However, if a valid copy exists on another replica, btrfs will
+transparently correct the read error, and running ``btrfs scrub`` in read-write
+mode will fix the error permanently by copying the valid metadata block over the
+invalid one. Otherwise, the recovery tool ``btrfs restore`` is able to ignore
+the errors and attempt to restore the data, but due to the inconsistency in the
+metadata, the restored data will need to be manually verified by the user.
 
 The root cause of the error cannot be easily determined, possible reasons are:
 
--- a/Documentation/ch-volume-management-intro.rst
+++ b/Documentation/ch-volume-management-intro.rst
@@ -116,3 +116,20 @@ In order to remove a device, you need to convert the profile in this case:
 
         $ btrfs balance start -mconvert=dup -dconvert=single /mnt
         $ btrfs device remove /dev/sda /mnt
+
+.. warning::
+   Do not run balance to convert from a profile with more redundancy to one with
+   less redundancy in order to remove a failing device from a filesystem.
+   As the name suggests, balance tries to balance data across devices.
+   Converting from e.g. raid1 to single may move data from the healthy device to
+   the failing device. This data will become irretrievable if the failing device
+   corrupts the new data or fails completely before ``btrfs device remove`` can
+   finish moving it back onto the healthy device.
+
+   To recover from a failing device with a replicated profile when you cannot
+   add enough new devices to maintain the required level of redundancy,
+   physically remove and replace the failing device, mount the filesystem with
+   ``-o degraded``, then use :command:`btrfs-replace` to replace the missing
+   device with the new one. Once the device is replaced, check
+   ``btrfs filesystem usage``, and if any single profiles are listed, run
+   ``btrfs balance start convert=raid1,soft`` to convert them back to raid1.
--- a/Documentation/index.rst
+++ b/Documentation/index.rst
@@ -91,6 +91,11 @@ is in the :doc:`manual pages<man-index>`.
 
    </td></tr></table>
 
+Need help?
+----------
+
+Assistance is available from the `#btrfs channel on Libera Chat <https://web.libera.chat/#btrfs>`_ or the `linux-btrfs mailing list <https://subspace.kernel.org/vger.kernel.org.html>`_. Issues with the userspace btrfs tools can be reported to the `btrfs-progs issue tracker on GitHub <https://github.com/kdave/btrfs-progs/issues>`_.
+
 .. raw:: html
 
    <hr />


Documentation/ch-scrub-intro.rst
```diff
@@ -6,14 +6,27 @@ On filesystems that use replicated block group profiles (e.g. raid1), scrub will
 also automatically repair any damage by default by copying verified good data
 from one of the other replicas.
 
+.. warning::
+   Files with the NOCOW (``chattr +C``) attribute imply ``nodatasum``.
+   Errors in these files cannot be detected or corrected by scrub. This means
+   the NOCOW attribute is not currently safe in the presence of replicas since
```
adam900710, line 12: I'd say the `not safe` claims is not ideal.
This means you're calling all file systems without data checksum unsafe.

And even for btrfs, there are known IO patterns (abuse) that can lead to mismatch data checksum directly.
Namely incorrectly handled writeback (modify the direct IO memory while it's still under writeback).

Documentation/ch-scrub-intro.rst
```diff
@@ -6,14 +6,27 @@ On filesystems that use replicated block group profiles (e.g. raid1), scrub will
...
+   btrfs cannot determine which copy of data to use after a failure and may
+   return data from any replica, including corrupt or outdated data.
+   Detecting and recovering from a failure in this case requires manual
+   intervention. See issue `#482 <https://github.com/kdave/btrfs-progs/issues/482>`_.
+   
+   Notably, `systemd sets NOCOW on journals by default <https://github.com/systemd/systemd/commit/11689d2a021d95a8447d938180e0962cd9439763>`_,
+   and `libvirt ≥ 6.6 sets NOCOW on storage pool directories by default <https://www.libvirt.org/news.html#v6-6-0-2020-08-02>`_.
+   Other distributions may also enable NOCOW on database files or directories to
```
adam900710, line 20: And it's not only users' choice, but certain IO pattern can lead to csum mismatch by themselves (although that counts as an abuse though).

adam900710 (changes requested):

Documentation/ch-scrub-intro.rst
csnover: Sure, let me make sure there’s a mutual understanding of both the purpose of this warning and the operation of btrfs so I can update the wording to make sure it is communicating the right thing without being too inaccurate.

The purpose of this warning is to prevent users from experiencing data loss when they think btrfs is protecting their files against single replica corruption. It is not clear right now that simply having a healthy replica is insufficient to avoid data loss when using NOCOW. For example, [this Reddit thread](https://old.reddit.com/r/btrfs/comments/xki69r/nodatacow_and_data_loss/).

In my case, I had one journald file end up corrupted, and did not understand how that was possible unless my *whole* filesystem was corrupted, until someone else pointed out that journald sets those files to NOCOW, and that is why *only* that file ended up being irreparably damaged.

With NOCOW, my understanding is that btrfs does not even meet standards of traditional raid1 systems which would have marked the unhealthy replica as tainted and only read from the healthy one, thus avoiding corruption in this scenario. As such, I pick this language to express that at least with single profile, your file might be silently corrupted if there is a crash, but you know that already because you have no replica. With replicas, it is “not safe” because having a copy committed to a healthy replica does *not* (apparently, unless it’s a bug) prevent corruption.

It is not clear to me how btrfs handles NOCOW exactly since e.g. #134 talks about multiple copies being possible to restore, but I would expect that only to be true until btrfs notices the metadata is inconsistent and fixes it up using the one with the correct checksum/transid and marks the blocks from the bad metadata as free again. (Clearly I do not understand the architecture of btrfs very much at all and am trying to potentially apply knowledge from other domains that do not apply, so please correct and explain so I can understand and be smarter :-))

adam900710: > With NOCOW, my understanding is that btrfs does not even meet standards of traditional raid1 systems which would have marked the unhealthy replica as tainted and only read from the healthy one, thus avoiding corruption in this scenario.

For NOCOW, the other RAID1 implementations go marking the bad device with extra flags, and I believe the btrfs part can indeed be improved.

But if the data corruption happens without really triggering any read error (e.g. for whatever reason the data on-disk just changed, without triggering a read error), then it's no better than NODATACOW btrfs RAID1, since just like NODATACOW btrfs, the block layer RAID1 won't detect any error, thus go the same path as btrfs (to go load balance the read)


That's the main reason I won't call it unsafe, because for above case the block layer RAID1 implementation is just as bad as btrfs.

However I also won't decline that btrfs is pretty bad at detecting unreliable disks, thus it won't avoid reading that bad disk.

I'd prefer to make it clear that, btrfs doesn't have any ability to detect bad disks (which can be very complex by itself, and I'm very curious how other RAID1 handles it), so for that NODATACOW + bad disks case, users will get data corruption, that's correct.

But it doesn't mean the other RAID1 is that safe, they just mask their lack of data csum by detecting if the disk is bad.

> It is not clear to me how btrfs handles NOCOW exactly since e.g. https://github.com/kdave/btrfs-progs/issues/134 talks about multiple copies being possible to restore, but I would expect that only to be true until btrfs notices the metadata is inconsistent and fixes it up using the one with the correct checksum/transid and marks the blocks from the bad metadata as free again.

First of all, NOCOW can only be applied to data, thus we always call it NODATACOW.
Metadata is *always* checksumed, and *always* COWed. So metadata is always safe.
(as long as the disks implement the barrier operations correctly)

And unfortunately even with RAID1C3/C4, btrfs won't compare all the mirrors and choose the correct one (the one matches the most mirrors), mostly because it will bring a huge performance drop.

It's back to the bad disks detection which is lacking in btrfs, unfortunately.

csnover: > For NOCOW, the other RAID1 implementations go marking the bad device with extra flags, and I believe the btrfs part can indeed be improved.

Yes, if some enhancement like this is done then I think btrfs raid1 will be guaranteed to never be worse than traditional software raid1 in all failure modes, which would be great.

> But if the data corruption happens without really triggering any read error (e.g. for whatever reason the data on-disk just changed, without triggering a read error), then it's no better than NODATACOW btrfs RAID1

Sure, this is why I personally opted to use btrfs raid1 instead of mdadm. :-) If I understand correctly, it is only the combination of NOCOW + media failure where btrfs ends up in a worse state due to what you mention about bad disk detection and then load balancing reads the wrong data. Unfortunately media failure is probably the most common failure mode. :-(

> just like NODATACOW btrfs, the block layer RAID1 won't detect any error, thus go the same path as btrfs (to go load balance the read)

So I think this is right, but just to confirm: if a user purges OS file cache and retries, eventually they will read the good data, since the metadata is kept consistent on both mirrors always, and the corruption is just due to how one disk did not write out the data blocks pointed by the metadata? This makes a lot of sense and is straightforward enough that I may provide this additional context in the documentation.

If I got that right, is it possible to force a read from one device (without mounting degraded with only the single healthy device connected)? It would be good to mention in the documentation if this is the case. It would be very sad if users are “losing” data in this situation only because they can’t tell btrfs to force read/overwrite the file from the known good disk. I understand #482 is a request to add something akin to this to scrub but wonder if it is possible to do it in an ugly way now?

I did not remove the reference to the bad journal file on my filesystem yet, so I wonder if I could test this. So far it seems like either the file got is cached, or I get unlucky and never hit the right PID to balance to the healthy device, or journald read the bad log on reboot, processed it into garbage, then deleted the original inode, so now it is actually a file that is just bad forever. I have no idea!

> I'd prefer to make it clear that, btrfs doesn't have any ability to detect bad disks (which can be very complex by itself, and I'm very curious how other RAID1 handles it), so for that NODATACOW + bad disks case, users will get data corruption, that's correct.

OK, I think I can rewrite to be more like this since that feels better to you and it seems OK to me.

> But it doesn't mean the other RAID1 is that safe, they just mask their lack of data csum by detecting if the disk is bad.

Yes, I understand. The “not safe” phrase was meant only by comparison to btrfs with CoW data mode, not to any others. :-)

> First of all, NOCOW can only be applied to data, thus we always call it NODATACOW.

Sure, this makes sense. I used the terminology NOCOW here because that is how I see the `+C` attribute described elsewhere (or else in the `chattr` man page, `No_COW`). I think it will be good to also clarify in the documentation so will try to work this in too.

> And unfortunately even with RAID1C3/C4, btrfs won't compare all the mirrors and choose the correct one (the one matches the most mirrors), mostly because it will bring a huge performance drop.

This also seems like a good information to write in the docs, just for anyone who wonders about it, so I will try to work that in too.

Changes should come tomorrow and I will drop a note when that happens.

adam900710: > if a user purges OS file cache and retries, eventually they will read the good data, since the metadata is kept consistent on all replicas always, and the corruption is just due to how one disk did not write out the data blocks pointed by the metadata? 

Firstly for metadata read, if we hit any bad (csum/generation/level mismatch), we immediately try the other mirrors. Either we got a good one, then writeback the good content to the bad mirror. Or we exhaust all copies and return error.

Then for the data, if it has checksum (and checksum is stored in metadata), we go the same way above for repair.
If it has no checksum, we just return the content we read out (which mirror is based on the pid based load balance).

So for data with checksum, we will always grab a good mirror if possible, and a read on the bad mirror will also cure the bad mirror (if there is a good one).

> If I got that right, is it possible to force a read from one device (without mounting degraded with only the single healthy device connected)?
Following my previous answer, if there is data checksum, no force needed, we immediately repair any bad data copy if we have any good mirror remaining.
(For metadata it's always repaired since it has csum/level/generation checks)

For data without csum, if you can get a good mirror really depends on the pid of the read process.
It's not as good as "forcing btrfs to read certain copy", but if you really want to go that path, it's still possible using "btrfs-map-logical" command to manually read certain mirror of a logical bytenr, and fetch the good copy.

But NODATACOW also means, we won't COW the data write, meaning if your process read out the bad content, modify some bytes, and write it back, the good copy is also gone now.

So for your journal case, I do not think it's a good chance to get the good mirror back, because later writes are definitely going to screw the good mirror.

Zygo: Currently btrfs doesn't have any way to reliably resync data between mirrors if a device goes offline and comes back while mounted.  Excluding recovery methods which require umounting the filesystem, the only tool we have for this case is scrub, and scrub's reliability for data blocks is limited to the csum collision rate (this isn't a problem for metadata blocks because there are other header fields to indicate a lost write, so a csum collision is not sufficient to create an undetected error in metadata).  For crc32c csums, that's an error rate of 1 in 16 TiB, high enough for multiple undetected errors per drive in the worst case on modern hardware.  For nodatacow files, the undetected error rate on the offline drive is always 100% of errors, because scrub doesn't do anything with nodatacow files at all.

In mdadm it's possible to explicitly mark a device as bad and then force a resync in place (i.e. replacing the drive with itself).  In this case, mdadm reads only from the good devices, so it restores correct data reliably to the bad ones (assuming there's only one corrupt drive and the user knows which one that is, but this is quite common when a single controller or cable fails).

It's possible to do resync in place in btrfs by patching `btrfs replace` to skip the mounted check, then running `btrfs replace start -r` through a loop device to avoid the kernel's same-device check.  We do this whenever we need to get one drive reliably resilvered after an online corruption event.  It would be nice if upstream `btrfs replace` supported this operation directly.

Currently the kernel code explicitly forbids a replace target from being the same device as the source,  and btrfs-progs userspace checks to see if the device is an existing device in the filesystem before allowing replace, without making an exception for self-replace.  Both of those are bugs that should be fixed, or scrub should be extended to support resilvering explicitly.

csnover: > For crc32c csums, that's an error rate of 1 in 16 TiB, high enough for multiple undetected errors per drive in the worst case on modern hardware.

I am curious about the worst case here. Would it be that all data blocks of a large filesystem are rewritten between device disappearing and reappearing? I would naïvely assume that the number of changed data blocks would typically be low enough that the chance of a collision on any actually changed block due to a transient dropout would be sufficiently low to not have to worry in practice, but I don’t really trust my judgment at all about anything.

Notwithstanding, the ability to *guarantee* a zero error rate by resilvering back certainly sounds preferable to me when there’s no need to take that risk, if that makes sense, and is what you are saying.

I don’t like the feeling I have had during this hardware event that btrfs volume management is optimised for making changes to healthy filesystems and unexpectedly difficult to reason about in a failure scenario. I’m not sure where is the best place to discuss this so maybe something can be improved.

csnover: I’ve added two new commits to this patch set, one that hopefully satisfies this feedback, and then a new warning in the section that talks about using balance to remove a device from a redundant set. As always I am not super confident that I am speaking accurately, so please check and let me know about any inaccuracy. Thanks!

adam900710 (approved):

adam900710:
The latest version looks good to me.

I'll still leave the PR open for any extra feedback, until it got merged.

Documentation/ch-scrub-intro.rst
```diff
@@ -7,18 +7,30 @@ also automatically repair any damage by default by copying verified good data
...
+   Setting the ``No_COW`` (``chattr +C``) attribute on a file implicitly enables
+   ``nodatasum``. This means that while metadata for these files continues to
+   be validated and corrected by scrub, the actual file data is not.
+
+   Furthermore, btrfs does not currently mark missing or failed disks as
+   unreliable, so will continue to load-balance reads to potentially damaged
+   replicas in a replicated filesystem. This is not a problem normally because
+   damage is detected by checksum validation and a mirror copy is used, but
+   because ``No_COW`` files are not protected by checksum, bad data may be
+   returned even if a good copy exists on another replica. Which replica is used
```
Zygo, line 19: Which replica is used is determined by the read policy, which can be changed through sysfs; however, the only currently implemented read policy in upstream Linux is 'pid'.

So this isn't wrong today, but one day it may be.

It might be better to say "The filesystem's configured read policy determines which replica will be used."

Documentation/ch-scrub-intro.rst
Zygo: > Would it be that all data blocks of a large filesystem are rewritten between device disappearing and reappearing?

Big corruption events do happen, but they are typically caused by failures between btrfs and the devices, not the devices themselves.  A couple of examples from earlier this year:

1.  An unmonitored filesystem (e.g. a home user with a NAS) might have a drive go offline that is not noticed until weeks later when it comes back online after a reboot.  Daily data rates these days can be large enough to get to double-digit TB in that time, and they're only going up.

2.  We recently had an incident where `pvmove` went badly wrong during a routine upgrade of an 18 TB LV device under a btrfs filesystem.  Everything seemed normal:  pvmove finished with no error, iostat reported a little over 18 TB read and 18 TB written during pvmove...but it turned out that only about 6 TB of data had actually been moved to the correct location.  This wasn't discovered until several hours into the post-pvmove scrub.  From the btrfs point of view, there were 12 TB of randomly corrupted data on the upgraded device.  We weren't going to roll the dice on a 75% chance of an undetected error with a scrub, so we aborted the scrub and forced a full device replace instead.

The risk for datasum files can be substantially reduced by using xxhash64 or larger instead of crc32c for csums, but this requires preparation in advance (best at mkfs time), as it's not convenient to change csum method on the fly.

> the ability to _guarantee_ a zero error rate by resilvering back certainly sounds preferable

For nodatacow files, it's not a matter of preference--resilvering or replacing the bad drive is the only recovery method that can work _at all._  (and it only works if the problem is confined to a single drive and the user knows which drive is affected)

> unexpectedly difficult to reason about in a failure scenario

It's easy to reason about failure scenarios if you're familiar with the implementation details, but that familiarity is also the _only_ way to reason about failure scenarios.  btrfs has unique architectural features, so knowledge of other filesystems or RAID schemes doesn't necessarily apply (e.g. in ext4 or xfs on mdadm or lvm, a user can't exclude an individual file from replication or dm-integrity, but on btrfs they can).  There are some things that can be improved, but also some things that are inevitable fallout of mutually exclusive goals.

Documentation/ch-scrub-intro.rst
csnover: Sure, I will clarify.

Documentation/ch-scrub-intro.rst
csnover: > It's easy to reason about failure scenarios if you're familiar with the implementation details, but that familiarity is also the only way to reason about failure scenarios.

Sure, some of it is the different design approach and that’s OK, especially since those differences mean that btrfs fixes many fundamental problems of other filesystems.

To me, the unexpected difficulty comes from how I think this particular user story shouldn’t require anyone to become a filesystem implementation expert: “I want my system to remain online and usable even if one of two disks fails, without any data corruption due to the failure. I want to be able to replace the failed disk using a single command (or two commands if partitioning cannot be automated), and have confidence that this one command will correctly restore everything to the intended level of redundancy. I do not want to feel confused about what steps are safe or not, nor inadvertently run commands that will non-obviously cause or increase the risk of data loss or downtime.”

In the case of mdadm, the failing disk would be flagged and write-intent bitmap would be used to mark what data didn’t make it to the failed disk. No corruption[*] or confusion about what steps to take. Just remove old disk, insert new disk, partition, run `mdadm --add /dev/X` or `mdadm --replace /dev/Y --with /dev/X`, and wait. You can even mark the failing disk to not be used at all in advance if it is causing problems. In any case, one hardly needs to know anything about the implementation, and the parts that need to be known are simple.

In the case of btrfs, it does not notice the disk failed at all, so starts using it again like normal after a reboot even though it is unhealthy. At least some `nodatasum` data gets lost probably immediately due to unlucky load balancing. Correction of remaining `datasum` data is apparently possible only by checksum correction, and now we talk about how probably that isn’t too safe either except for metadata since btrfs default is crc32c which has very low collision resistance?

The actual recovery requires either adding an extra disk first, or using `-o degraded`. If adding an extra disk first, when running `btrfs replace`, the default behaviour is not with `-r` flag, so if there is any bad `nodatasum` data on the failing disk or `datasum` data where there was a hash collision, by default, corrupted data will be written to the new replacement, instead of good data from the healthy device. (Does `-r` absolutely guarantee no reads from the unhealthy device in this case, or is it mandatory to pull the unhealthy device and use the `-o degraded` path?)

If using `-o degraded` on a root filesystem, it probably also needs to be done from a recovery environment, since setting `-o degraded` in advance sounds like one good way to accidentally get a split brain if there is ever any number of other types of fault that cause the healthy disk to temporarily not appear. There’s also conflicting information on the internet about whether this mount option is a risky one-chance operation, or if that was just a bug that got fixed (official docs don’t seem to say one way or the other). After getting to this point, the user then needs to know what devid to use since the old disk won’t exist to replace; according to [this wiki](https://wiki.tnonline.net/w/Btrfs/Replacing_a_disk) (which contains more important-seeming warnings that don‘t seem to be listed in the official documentation), the user has to run `btrfs device usage` to find the missing disk ID, then put that into `btrfs replace`.

In both cases, the user must remember afterward not to forget to run `btrfs balance`, or else now some of their data is a time bomb.

Especially as btrfs becomes more mainstream and takes over as the default filesystem for some distributions, I really feel like it needs to be about as simple, and at least as reliable, as mdadm in this failure scenario. I don’t see why there would not be a path to get there, but some of these problems seem to have been raised [in at least 2018](https://www.mail-archive.com/linux-btrfs@vger.kernel.org/msg78254.html), so it feels bad to encounter the same thing six years later.

OK, thanks for listening to my TED talk. I think everything I wrote is accurate but as usual please correct any misconceptions or, if you would, unanswered questions. My apologies if this comes across as hostile; again, I think btrfs is doing more right things than wrong things, and so when btrfs makes most things in daily operation quite easy, the unexpected difficulty of a disaster recovery is heightened. The double-edged sword of success :-)

[*] Pedantic hedging here: I don’t know this for sure, I have never personally experienced this situation, maybe it breaks too, but all evidence suggests it would do the right thing.

Documentation/ch-scrub-intro.rst
csnover: I’ve reworded this to mention the read policy setting.

Documentation/ch-scrub-intro.rst
Forza-tng: >  chattr -C result is “Operation not permitted” on this file and three others, not sure if that’s intended or why it would be restricted, no output anywhere says why.

cow/nocow mode can not be toggled on files that have data. The man page for `chattr` states: 
> Note:  For btrfs, the ’C’ flag should be set on new or empty files.

csnover: >  The man page for `chattr` states: […]

Sure, but I was trying to *clear* the flag, not set it, you see :-)

By the way, thank you for your wiki content, I hope that much of it makes its way into the official documentation someday; the examples for how to handle disk replacement were the best instructions I found and I think the official documentation deserves that level of clarity.

Forza-tng: > > The man page for `chattr` states: […]
> 
> Sure, but I was trying to _clear_ the flag, not set it, you see :-)

True, we could always improve documentation and be better at providing good UX for users.

> By the way, thank you for your wiki content, I hope that much of it makes its way into the official documentation someday; it is quite well organised and the examples for how to handle disk replacement were the best instructions I found and I think the official documentation deserves that level of clarity.

Thanks! I am glad the wiki is useful. I started it mostly as a way to learn Btrfs and write down how to do things in an easy way. It's great to see so much improvement that has gone into the official documentation since then 🙏

> With NOCOW, my understanding is that btrfs does not even meet standards of traditional raid1 systems which would have marked the unhealthy replica as tainted and only read from the healthy one.

IMHO, it may not always be optimal to mark a disk as bad when it clearly can read some data because it could happen that data on the good disk also becomes corrupt. Still, some logic to determine how and when a disk should be dropped/marked bad would be useful. Probably best implemented as a user-space daemon that monitors error rate, etc. 

Also, to clarify that md/dm raid also would not detect corruption. They would detect read-errors, but that would Btrfs too. 

> And unfortunately even with RAID1C3/C4, btrfs won't compare all the mirrors and choose the correct one (the one matches the most mirrors), mostly because it will bring a huge performance drop.

Such a feature could probably be added to scrub. It could let users detect and choose which copy to preserve. https://github.com/kdave/btrfs-progs/issues/482

csnover: > IMHO, it may not always be optimal to mark a disk as bad when it clearly can read some data because it could happen that data on the good disk also becomes corrupt. Still, some logic to determine how and when a disk should be dropped/marked bad would be useful. Probably best implemented as a user-space daemon that monitors error rate, etc.

Sure, I can see arguments on both sides of this, but it can only really be decided once btrfs has a functioning write-intent bitmap. Reads from a failed device—which mdadm will do during recovery when the failed device is still accessible to avoid stressing other disks into failure—are only safe with a write-intent bitmap. It is unsound for btrfs to continue reading from a disk with failed write because it has no write-intent bitmap to skip/fix outdated blocks. This is absolutely true for nodatasum data, but it is also true even for datasum data due to possibility of hash collision, as mentioned by @Zygo—though the risk is low enough *in practice* that you can probably get away with it a lot of the time (I managed).

> Also, to clarify that md/dm raid also would not detect corruption. They would detect read-errors, but that would Btrfs too.

mdadm maintains write-intent bitmap, so does detect one form of “corruption” that btrfs does not, which is “corruption” due to writes that only made it to one device.

I tried to drive it home on my mailing list post, and I try to do the same here: btrfs needs write-intent bitmap urgently. Without write-intent bitmap, btrfs is like a super-advanced car with the most advanced airbags and collision-avoidance sensor technology, but no seatbelts.

Documentation/ch-scrub-intro.rst
zatricky: I feel like this "feature" is a huge failure on the part of the libvirt developers. Checksums are super important for data integrity and they are disabled by this "default behaviour". For me to only find out about it after 4 years thanks to this PR feels embarrassing.

I checked all my servers that use btrfs and found that most did not have this problem - but that is likely because the image subvolumes and files were created via automation before the storage pool configurations were created. However, whenever manual action was taken, there was always a chance that the storage pool configuration would create the path with nocow.

csnover:
Why is this not merged yet?

adam900710:
This is conflicting with the latest devel branch.

Mind to rebase the branch to the latest devel?

csnover:
This PR was broken by whatever happened at https://github.com/kdave/btrfs-progs/pull/863#commits-pushed-bc39df6 that caused extra stuff to be added to the branch. I ran `git rebase --skip` over these and force pushed.

csnover:
CI is failing because base branch is broken, not due to changes in this PR.

Documentation/trouble-index.rst
```diff
@@ -21,17 +22,20 @@ contains target block offset and generation that last changed this block. The
...
-cannot be easily recovered, because of information loss. The recovery tool
-``btrfs restore`` is able to ignore the errors and attempt to restore the data
-but due to the inconsistency in the metadata the data need to be verified by the
-user.
+Once the mismatched generations are stored on the device, without a good copy
+from another replica, it's permanent and cannot be easily recovered because of
+information loss. However, if a valid copy exists on another replica, btrfs will
+transparently correct the read error, and running ``btrfs scrub`` in read-write
+mode will fix the error permanently by copying the valid metadata block over the
+invalid one. Otherwise, the recovery tool ``btrfs restore`` is able to ignore
```
adam900710, line 36: Nowadays we have `-o rescue=all,ro` mount option, which is a more useful solution compared to `btrfs restore`

Thus we may want to promote the new solution, and use `btrfs restore` as the last option (and under most cases it's not that better compared to rescue=all)

Documentation/btrfs-check.rst
```diff
@@ -19,6 +19,12 @@ by the option *--readonly*.
 
 :command:`btrfsck` is an alias of :command:`btrfs check` command and is now deprecated.
 
+.. note::
+   Even though the filesystem checker requires a device argument, it scans for all
+   devices belonging to the same filesystem and may report metadata errors from other
+   devices that are correctable by :command:`btrfs scrub`. In this case, run scrub
+   first to ensure any correctable metadata errors are fixed to avoid false-positives.
```
adam900710, line 26: I do not think this note is needed.

Btrfs check handles metadata error just like kernel, if a mirror is corrupted but still have another good one available, btrfs check won't report it as an error. E.g:

```
Opening filesystem to check...
checksum verify failed on 30490624 wanted 0xcdcdcdcd found 0x86f5c8d8 << Mirror 1 corrupted for csum root
Checking filesystem on test.img
UUID: f01a583a-df1e-414b-a24c-7fe8bf2ef019
[1/8] checking log skipped (none written)
[2/8] checking root items
[3/8] checking extents
[4/8] checking free space tree
[5/8] checking fs roots
[6/8] checking only csums items (without verifying data)
[7/8] checking root refs
[8/8] checking quota groups skipped (not enabled on this FS)
found 147456 bytes used, no error found <<< Still no error.
total csum bytes: 0
total tree bytes: 147456
total fs tree bytes: 32768
total extent tree bytes: 16384
btree space waste bytes: 140595
file data blocks allocated: 0
 referenced 0
```

So there is no false positive. And `btrfs scrub` has a much higher bar to clear (RW mount), meanwhile btrfs check is always the most comprehensive check tool.

csnover: This directly contradicts my experience when using btrfs-progs 6.6.3. In which version was this changed/fixed?

adam900710: IIRC it's always the case from the beginning.

Mind to give an example where btrfs-check is reporting recoverable corrupted mirrors as an critical error?
You can use `btrfs-map-logical` to find out the physical address of a metadata copy and corrupted it.

csnover: The history of the problem is linked in the PR description. Briefly: a disk failed and temporarily dropped from the array. Therefore, the two disks were no longer in sync after rebooting. `btrfs check` claimed the filesystem was corrupted and printed messages to that end.

In consultation with @Zygo on IRC I manually verified by `btrfs inspect-internals` that `btrfs check` was reporting invalid node types/values that existed only on the bad disk. The kernel logs showed the kernel driver was noticing the invalid metadata nodes on the bad disk and switching to the good one, but `btrfs check` was claiming filesystem corruption even though the good disk was not corrupted and the filesystem metadata was fine. There was no indication by `btrfs check` that it was checking (only? due to PID load-balancing?) the bad device even when explicitly given the path to the good disk. Only after read-write `btrfs scrub` fixed up the metadata on the bad disk did `btrfs check` report no errors.

adam900710: Do you still have that fs?

I'm checking the `btrfs_read_extent_buffer()` function and still can not find out a good way where split-brain case can make btrfs-check to go the bad copy without trying the good one.

I think it's more like a bug other than an expected behavior.

csnover: > Do you still have that fs?

I do not. However, until something is done to fix the lost writes problem (like [this MVP idea from Oct 2](https://lore.kernel.org/linux-btrfs/b5b3df31-627a-403c-af4d-7495e7c28bad@zetafleet.com/)), it seems like a synthetic reproducer should be trivial: create a raid1 fs from two disks, forcibly remove one disk, ensure btrfs continues writing to the remaining disk for a little bit, then unmount the fs, reconnect the forcibly removed disk, and remount.

> I'm checking the `btrfs_read_extent_buffer()` function and still can not find out a good way where split-brain case can make btrfs-check to go the bad copy without trying the good one.

If the problem can’t be found in the latest version then maybe it was fixed between 6.6.3 and current head? This was btrfs-progs 6.6.3 (which is [still the latest available packaged version for Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1074237)) on kernel 6.9.10.

> I think it's more like a bug other than an expected behavior.

Sure, this seems likely to me.

adam900710:
Although I have merged this series, I'm going to do a lot of updates on the words and of-course add SoB lines to all involved patches, re-order patches (some patches are fixing things of previous one, which should not happen for out-of-tree patches).

I'm a little too optimistic of this merge, and github web is not a good way to review patches, it hides the SoB line of the author, thus makes it harder to find a missing one.

I'll let you informed when the devel branch is updated so you can verify if the modification is fine.

adam900710:
The branch is now updated (`devel` branch). I have done the following modifications:

- Add an SoB line for all patches
  I'm using `<csnover@users.noreply.github.com>` as your mail address.
  If you want to use a specified address, please let me know.

- Merge all scrub related commits into a single one

- Remove btrfs-check related warnings
  As I explained, btrfs-check is designed to choose the good copy (for metadata at least) just like kernel.
  So the warnings that bad mirror will cause false alerts is at least not the expected behavior

- Explain scrub in more details
  It's done on a per-device basis, and full fs scrub is done by scrubbing each device in parallel.

  Furthermore the behavior of scrub repair (using a good mirror) is also done for all the metadata/data reads,
  There is really nothing that special about scrub, thus a lot of recommendation to run `btrfs scrub` before btrfs-check
  is removed.

- Remove the pid based load balance patch
  It's at the bad timing where we're also implementing other read policies.
  I'd prefer to remove all the pid based read balance mention for now.

csnover:
To get the easier administrative part out of the way: I didn’t actually sign off on whatever was changed, so you probably shouldn’t list me in such a field, but I also don’t really care, so using that email address is totally fine if you need the field to be there to get these changes published.

Regarding the other feedback: I’m starting to wonder if there is an irreconcilable difference in philosophy here, but I will try advocating for my position one last time in the hope of at least achieving a mutual understanding of why I don’t think the outright removal of some of the notes I added is OK.

First, please know that it is not my intent to attack anyone or their work personally. I understand the many limitations caused by the under-resourced and voluntary nature of OSS work. If my feedback ever feels overly harsh or critical, it is because of my conviction that this is extremely important to get across, and so I am speaking very directly.

So: What is the purpose of end-user documentation? Is it supposed to help users to understand and operate the system safely as it actually is, or is it supposed to be a white paper that describes an idealised future version of the system? I believe it to be the former. The feedback I’m receiving on this ticket suggests a belief that it is closer to the latter.

Just because something will be changed in the future doesn’t mean its current or past behaviour should be left undocumented. btrfs is not an evergreen SaaS. People will be using the old versions for years and they need to know how those old versions work too. If the behaviour changes, then the documentation changes to read “in versions before N”. Descriptions of old behaviour can be moved to an “obsolete” page when N-1 is not still in supported LTS releases, but retaining such information somewhere official is necessary to clarify whatever outdated information is still floating out there on the internet.

Similarly, if the purpose is to help understanding and safe operation, then just because something is a bug doesn’t mean it should not be documented. There is even an industry term for this: errata. Even when a bug is self-evident (e.g. a crash), it still warrants documenting if there is a known workaround. In the case of `btrfs check` returning false positives, there are two reasons why I find it more critical to document than a run-of-the-mill defect:

1. `btrfs check` *appears* to be working when it is not, thus it is not self-evident there is a bug causing false positives; and
2. The need for clarity is heightened in high-stress situations like potential data loss/disaster recovery scenarios.

Whether the behaviour is intended is irrelevant to what the end user needs to know to understand and operate the system, and such information needs to be written down somewhere, in a place where end-users will think to look, which is official and up-to-date.

By way of suggested alternatives: if the btrfs bug tracker wasn’t such an unkempt graveyard, I’d be more receptive to the argument that tickets provide a reasonable stand-in for errata. But, as it stands, there are at least three separate bug trackers I’m aware of (and I am guessing SUSE has an internal one, which would make four); the kernel tracker alone has 296 open bugs, 97% of are still in NEW status. Many of them, like “Add option to check progress/status of btrfs-replace job”, are clearly obsolete. If no one is doing the minimum necessary work of project management to ensure bug trackers are reliable sources of information, they simply can’t be used as alternatives to documentation.

To look at this another way: Who is *harmed* by these notes? Do they suggest performing a dangerous operation? Do they enhance, or harm, the user’s understanding? Do they overwhelm other more important details, and if so, how can the same information be presented in a way that doesn’t do that, rather than just omitting unpleasant reality?

Given the inability to immediately find a code path that could be causing incorrect diagnostics from `btrfs check` and the false nature of many user reports of issues, I recognise the scepticism about my report that `btrfs check` did report errors on an array with only one corrupted mirror, which went away after `btrfs scrub` corrected the errors on the bad mirror. I don’t know what to say other than to fall back on my reputation as not being the kind of person who is generally incapable of telling whether a command is printing out errors or not. Trying to run `btrfs check` in lowmem mode even segfaulted! Surely that wouldn’t happen if the metadata it was reading was from the good mirror?

Let me know your thoughts, if you’d like.

Best,

adam900710:
I think you miss one point, that an RW mount on a possible corrupted can cause problem by itself.
E.g. for a corrupted extent tree/free space cache/free space tree, which breaks COW, just RW mounting with the extra update on the atime or whatever can already do damage further.

Thus I do not recommend RW scrub as a way to fix things, as it has the possibility to further corrupt the filesystem.


Except certain out-of-sync devices, the common corruption like mismatch transid can not really be repaired by using scrub. It's more possible some device are not honoring FLUSH/FUA correctly, and that will be a bigger problem by itself.

Furthermore if scrub can fix it, then there should be no observable problem except the dmesg error messages.
As I have explained in the updated doc, kernel data/metadata read itself will do the repair, and we have several test cases to verify that behavior (including for scrub and for regular read).

So the idea of running RW scrub to self-heal is not any better than just running the filesystem as usual.

PR #866, opened 2024-08-08
Remove stray newline in filesystem show

mattlangford:
This PR removes a stray newline in this command:
```
Label: 'data'  uuid: b84e1673-5d07-4ad2-a43c-abb0aa3ff0cc
	Total devices 2 FS bytes used 413.91GiB
	devid    1 size 860.47GiB used 447.03GiB path /dev/nvme0n1p5
	devid    2 size 931.32GiB used 447.03GiB path /dev/sda2

Label: 'testing'  uuid: f16b4f19-3461-47d7-b9d2-91e452f39383
	Total devices 1 FS bytes used 144.00KiB
	devid    1 size 29.30GiB used 536.00MiB path /dev/sdb

```

Each prior `pr_verbose` already includes a newline, so it doesn't appear to be needed. I've validated this patch removes that newline at the end (but keeps newlines between filesystem entries) when running `btrfs filesystem show` locally.


kdave:
The newlines separate listings of multiple filesystems, with this patch it's all in one long text and not easy to visually skim just one filesystem. Showing a single filesystem does have a stray newline so this can be fixed but I'd rather keep the visual style of multiple filesystems.

mattlangford:
Makes sense to me! Do you think a refactor here to drop the newline on the last filesystem would be reasonable?

kdave:
Yes, this looks good. There are some minor issues, I'll add the review comments.

cmds/filesystem.c
```diff
@@ -479,6 +479,11 @@ static int btrfs_scan_kernel(void *search, unsigned unit_mode)
 
 		fd = open(mnt->mnt_dir, O_RDONLY);
 		if ((fd != -1) && !get_df(fd, &space_info_arg)) {
+			/* put spacing between filesystem entries for readability */
+			if (found != 0) {
+				pr_verbose(LOG_DEFAULT, "\n");
+			}
```
kdave, line 485: For single statements it's not necessary to use { }, which is the preferred coding style.

cmds/filesystem.c
```diff
@@ -755,6 +760,7 @@ static int cmd_filesystem_show(const struct cmd_struct *cmd,
 	char uuid_buf[BTRFS_UUID_UNPARSED_SIZE];
 	unsigned unit_mode;
 	int found = 0;
+	int needs_newline = 0;
```
kdave, line 763: Please use type bool.

cmds/filesystem.c
```diff
@@ -880,8 +889,14 @@ static int cmd_filesystem_show(const struct cmd_struct *cmd,
 		goto out;
 	}
 
-	list_for_each_entry(fs_devices, &all_uuids, fs_list)
+	list_for_each_entry(fs_devices, &all_uuids, fs_list) {
+		/* put spacing between filesystem entries for readability */
+		if (needs_newline) {
+			pr_verbose(LOG_DEFAULT, "\n");
+		}
```
kdave, line 896: Same here { }

kdave (changes requested):

cmds/filesystem.c
mattlangford: Got it, updated!

cmds/filesystem.c
mattlangford: Good call

kdave:
Added to devel, thanks.

PR #868, opened 2024-08-13
btrfs-progs: add --subvol option to mkfs.btrfs

maharmstone:
This patch adds a --subvol option, which tells mkfs.btrfs to create the specified directories as subvolumes.

Given a populated directory img, the command

$ mkfs.btrfs --rootdir img --subvol usr --subvol home --subvol home/username /dev/loop0

will create subvolumes usr and home within the FS root, and subvolume username within the home subvolume. It will fail if any of the directories do not yet exist.

---

This is version 6 of the patch; I figured it'd be easier doing things here rather on the mailing list, if we're using both.

I've incorporated Qu's suggestion that we can free the subvol list as we go along, and thus speed up iteration slightly. I've also changed it so that the --subvol value is again relative to the --rootdir rather than the working directory, which was this way because of a potential future patch allowing out-of-tree subvolumes.
I've discussed this with @DaanDeMeyer, who wants this for mkosi, and agreed that the best way to do "out-of-tree" subvolumes would be for a future --default-subvol option instead.

### DIFF
--- a/mkfs/main.c
+++ b/mkfs/main.c
@@ -440,6 +440,7 @@ static const char * const mkfs_usage[] = {
 	"Creation:",
 	OPTLINE("-b|--byte-count SIZE", "set size of each device to SIZE (filesystem size is sum of all device sizes)"),
 	OPTLINE("-r|--rootdir DIR", "copy files from DIR to the image root directory"),
+	OPTLINE("-u|--subvol SUBDIR", "create SUBDIR as subvolume rather than normal directory"),
 	OPTLINE("--shrink", "(with --rootdir) shrink the filled filesystem to minimal size"),
 	OPTLINE("-K|--nodiscard", "do not perform whole device TRIM"),
 	OPTLINE("-f|--force", "force overwrite of existing filesystem"),
@@ -1055,6 +1056,9 @@ int BOX_MAIN(mkfs)(int argc, char **argv)
 	char *label = NULL;
 	int nr_global_roots = sysconf(_SC_NPROCESSORS_ONLN);
 	char *source_dir = NULL;
+	size_t source_dir_len = 0;
+	struct rootdir_subvol *rds;
+	LIST_HEAD(subvols);
 
 	cpu_detect_flags();
 	hash_init_accel();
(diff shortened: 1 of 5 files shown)


maharmstone:
Whoops, fixed test

DaanDeMeyer:
I tested this https://github.com/systemd/systemd/pull/33498 and it works perfectly

mkfs/main.c
```diff
@@ -440,6 +440,7 @@ static const char * const mkfs_usage[] = {
 	"Creation:",
 	OPTLINE("-b|--byte-count SIZE", "set size of each device to SIZE (filesystem size is sum of all device sizes)"),
 	OPTLINE("-r|--rootdir DIR", "copy files from DIR to the image root directory"),
+	OPTLINE("-u|--subvol SUBDIR", "create SUBDIR as subvolume rather than normal directory"),
```
adam900710, line 443: I didn't see any man page updates, as it would be much better to have more detailed requirements for the options (like SUBDIR must be a relative path inside ROOTDIR, etc).

maharmstone: Thanks Qu. Pushed new version including man page change

maharmstone:
New version including adding --subvol to mkfs.btrfs man page

adam900710 (approved):
Looks good to me.

adam900710:
>I've discussed this with @DaanDeMeyer, who wants this for mkosi, and agreed that the best way to do "out-of-tree" subvolumes would be for a future --default-subvol option instead.

Just a little curious about this future feature. Do you plan to add a new option "--default-subvolume" or something like that, so that you can specify a subvolume *inside* the rootdir as the default one, after filling the rootdir along with the subvolumes?

If so, that new "--default-subvolume" option should still rely on the current rootdir + subvolume mechanism, just a new offline way to set default subvolue.

And if that's the case, everything looks fine to me, including the future feature.

maharmstone:
> > I've discussed this with @DaanDeMeyer, who wants this for mkosi, and agreed that the best way to do "out-of-tree" subvolumes would be for a future --default-subvol option instead.
> 
> Just a little curious about this future feature. Do you plan to add a new option "--default-subvolume" or something like that, so that you can specify a subvolume _inside_ the rootdir as the default one, after filling the rootdir along with the subvolumes?
> 
> If so, that new "--default-subvolume" option should still rely on the current rootdir + subvolume mechanism, just a new offline way to set default subvolue.
> 
> And if that's the case, everything looks fine to me, including the future feature.

Yes, that's right

Conan-Kudo:
FYI @schaefi, we probably want to use this in kiwi too!

PR #871, opened 2024-08-14
btrfs-progs: doc: add an extra note on --rootdir hard links for mkfs

adam900710:
Since commit cd6b81f4322f ("btrfs-progs: mkfs: rework how we traverse rootdir"), the --rootdir option has been reworked to get rid of the broken hard link creation.

Instead each hard link will be created as an independent file, and a runtime warning is added in commit f39b4deae6a3 ("btrfs-progs: rootdir: warn about hard links").

So here we also add an explicit note to the man page.


adam900710:
This is no longer needed as I'm just going to finish the hardlink detection/creation code.

PR #872, opened 2024-08-14
btrfs-progs: add --default-subvol option to mkfs.btrfs

maharmstone:
Adds a --default-subvol option to mkfs.btrfs, which works the same way as --subvol but also marks the subvolume as the default.

### DIFF
--- a/Documentation/mkfs.btrfs.rst
+++ b/Documentation/mkfs.btrfs.rst
@@ -160,6 +160,10 @@ OPTIONS
         directory.  The option *--rootdir* must also be specified, and *subdir* must be an
         existing subdirectory within it.  This option can be specified multiple times.
 
+--default-subvol <subdir>
+        As *--subvol*, except that it also sets the subvolume as the default for the
+        filesystem.  This option can only be specified once.
+
 --shrink
         Shrink the filesystem to its minimal size, only works with *--rootdir* option.
 
--- a/mkfs/rootdir.c
+++ b/mkfs/rootdir.c
@@ -99,6 +99,7 @@ static u64 g_hardlink_count;
 static struct btrfs_trans_handle *g_trans = NULL;
 static struct list_head *g_subvols;
 static u64 next_subvol_id = BTRFS_FIRST_FREE_OBJECTID;
+static u64 default_subvol_id;
 
 static inline struct inode_entry *rootdir_path_last(struct rootdir_path *path)
 {
@@ -436,6 +437,9 @@ static int ftw_add_subvol(const char *full_path, const struct stat *st,
 		return ret;
 	}
 
+	if (subvol->is_default)
+		default_subvol_id = subvol_id;
+
 	key.objectid = subvol_id;
 	key.type = BTRFS_ROOT_ITEM_KEY;
 	key.offset = (u64)-1;
@@ -701,6 +705,47 @@ static int ftw_add_inode(const char *full_path, const struct stat *st,
 	return 0;
 };
 
+static int set_default_subvolume(struct btrfs_trans_handle *trans)
+{
+	struct btrfs_path path = { 0 };
+	struct btrfs_dir_item *di;
+	struct btrfs_key location;
+	struct extent_buffer *leaf;
+	struct btrfs_disk_key disk_key;
+	u64 features;
+
+	di = btrfs_lookup_dir_item(trans, trans->fs_info->tree_root, &path,
+				   btrfs_super_root_dir(trans->fs_info->super_copy),
+				   "default", 7, 1);
+	if (IS_ERR_OR_NULL(di)) {
+		btrfs_release_path(&path);
+
+		if (di)
+			return PTR_ERR(di);
+		else
+			return -ENOENT;
+	}
+
+	leaf = path.nodes[0];
+
+	location.objectid = default_subvol_id;
+	location.type = BTRFS_ROOT_ITEM_KEY;
+	location.offset = 0;
+
+	btrfs_cpu_key_to_disk(&disk_key, &location);
+	btrfs_set_dir_item_key(leaf, di, &disk_key);
+
+	btrfs_mark_buffer_dirty(leaf);
+
+	btrfs_release_path(&path);
+
+	features = btrfs_super_incompat_flags(trans->fs_info->super_copy);
+	features |= BTRFS_FEATURE_INCOMPAT_DEFAULT_SUBVOL;
+	btrfs_set_super_incompat_flags(trans->fs_info->super_copy, features);
+
+	return 0;
+}
+
 int btrfs_mkfs_fill_dir(struct btrfs_trans_handle *trans, const char *source_dir,
 			struct btrfs_root *root, struct list_head *subvols)
 {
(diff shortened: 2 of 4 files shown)


maharmstone:
Fixed minor inelegance in `btrfs_set_default_subvolume` (no need to check return value at end of function)

DaanDeMeyer:
```
mkfs.btrfs: unrecognized option '--default-subvol'
usage: mkfs.btrfs [options] <dev> [<dev...>]

    Create a BTRFS filesystem on a device or multiple devices

    Allocation profiles:
    -d|--data PROFILE         data profile, raid0, raid1, raid1c3, raid1c4, raid5, raid6, raid10, dup or single
    -m|--metadata PROFILE     metadata profile, values like for data profile 
    -M|--mixed                mix metadata and data together 
    Features:
    --csum TYPE               
    --checksum TYPE           checksum algorithm to use, crc32c (default), xxhash, sha256, blake2 
    -n|--nodesize SIZE        size of btree nodes 
    -s|--sectorsize SIZE      data block size (may not be mountable by current kernel) 
    -O|--features LIST        comma separated list of filesystem features (use '-O list-all' to list features) 
    -L|--label LABEL          set the filesystem label 
    -U|--uuid UUID            specify the filesystem UUID (must be unique for a filesystem with multiple devices)
    --device-uuid UUID        Specify the filesystem device UUID (a.k.a sub-uuid) (for single device filesystem
                              only) 
    Creation:
    -b|--byte-count SIZE      set size of each device to SIZE (filesystem size is sum of all device sizes) 
    -r|--rootdir DIR          copy files from DIR to the image root directory 
    -u|--subvol SUBDIR        create SUBDIR as subvolume rather than normal directory 
    -D|--default-subvol SUBDIR
                              create SUBDIR as subvolume rather than normal directory, and set it as the default 
```

maharmstone:
> ```
> mkfs.btrfs: unrecognized option '--default-subvol'
> usage: mkfs.btrfs [options] <dev> [<dev...>]
> 
>     Create a BTRFS filesystem on a device or multiple devices
> 
>     Allocation profiles:
>     -d|--data PROFILE         data profile, raid0, raid1, raid1c3, raid1c4, raid5, raid6, raid10, dup or single
>     -m|--metadata PROFILE     metadata profile, values like for data profile 
>     -M|--mixed                mix metadata and data together 
>     Features:
>     --csum TYPE               
>     --checksum TYPE           checksum algorithm to use, crc32c (default), xxhash, sha256, blake2 
>     -n|--nodesize SIZE        size of btree nodes 
>     -s|--sectorsize SIZE      data block size (may not be mountable by current kernel) 
>     -O|--features LIST        comma separated list of filesystem features (use '-O list-all' to list features) 
>     -L|--label LABEL          set the filesystem label 
>     -U|--uuid UUID            specify the filesystem UUID (must be unique for a filesystem with multiple devices)
>     --device-uuid UUID        Specify the filesystem device UUID (a.k.a sub-uuid) (for single device filesystem
>                               only) 
>     Creation:
>     -b|--byte-count SIZE      set size of each device to SIZE (filesystem size is sum of all device sizes) 
>     -r|--rootdir DIR          copy files from DIR to the image root directory 
>     -u|--subvol SUBDIR        create SUBDIR as subvolume rather than normal directory 
>     -D|--default-subvol SUBDIR
>                               create SUBDIR as subvolume rather than normal directory, and set it as the default 
> ```

Oops, I forgot to add the long form to the list. Sorry Daan, try it now

DaanDeMeyer:
@adam900710 I am getting swarmed with hardlink warnings since your refactoring: https://paste.centos.org/view/a3a9b543. Any chance we can log this warning only for the first file instead of for every file?

kdave:
I'll update so the warning is printed first time and after the whole directory is processed so it's not lost.

adam900710:
> @adam900710 I am getting swarmed with hardlink warnings since your refactoring: https://paste.centos.org/view/a3a9b543. Any chance we can log this warning only for the first file instead of for every file?

Unfortunately if it's only warned once, it will mask all the other files.

Another thing is, although the timezone ones are indeed hardlinks, the python ones in my system doesn't has any hardlink.
Not 100% sure if the split time zone files will cause anything wrong.

I'm afraid we may need to implement hard link support sooner than later.

Documentation/mkfs.btrfs.rst
```diff
@@ -160,6 +160,10 @@ OPTIONS
         directory.  The option *--rootdir* must also be specified, and *subdir* must be an
         existing subdirectory within it.  This option can be specified multiple times.
 
+-D|--default-subvol <subdir>
```
adam900710, line 163: Not sure if we want a short option for this.
Especially this option is dependent on both `--rootdir` and `--subvolume`

mkfs/rootdir.c
```diff
@@ -95,6 +95,7 @@ static struct rootdir_path current_path = {
 
 static struct btrfs_trans_handle *g_trans = NULL;
 static struct list_head *g_subvols;
+static u64 *g_default_subvol_id;
```
adam900710, line 98: Can't we just make it a u64, not a pointer to u64?

And since we have such a value, why not directly set it inside `btrfs_mkfs_fill_dir()` and we can avoid the extra `default_subvolid` parameter?

adam900710 (commented):
Overall looks good to me.

Just some small nitpicks.

maharmstone:
Thanks Qu. I've force-pushed a version with your changes, and rebased it against 6.10.1.

adam900710 (approved):
Looks good to me.

mkfs/rootdir.c
```diff
@@ -701,6 +705,47 @@ static int ftw_add_inode(const char *full_path, const struct stat *st,
...
+	struct btrfs_dir_item *di;
+	struct btrfs_key location;
+	struct extent_buffer *leaf;
+	struct btrfs_disk_key disk_key;
+	u64 features;
+
+	di = btrfs_lookup_dir_item(trans, trans->fs_info->tree_root, &path,
+				   btrfs_super_root_dir(trans->fs_info->super_copy),
+				   "default", 7, 1);
+	if (!di || IS_ERR(di)) {
```
kdave, line 720: You can use IS_ERR_OR_NULL here

Documentation/mkfs.btrfs.rst
```diff
@@ -160,6 +160,10 @@ OPTIONS
         directory.  The option *--rootdir* must also be specified, and *subdir* must be an
         existing subdirectory within it.  This option can be specified multiple times.
 
+--default-subvol <subdir>
+        As *--subvol*, except that it also sets the subvolume as the default for the
+        filesystem.  This option can only be specified once.
```
kdave, line 165: Just a comment, this semantic sounds ok. Alternatively it could be --subvol DIR and a separate --default-subvol DIR that will only set the default bit, but it's more convenient to do it in one go.

kdave (approved):

mkfs/rootdir.c
maharmstone: Thanks, I've pushed a version with this change.

Documentation/mkfs.btrfs.rst
maharmstone: Yes, somebody else also suggested that, but I couldn't really see what the benefit was.

kdave:
Merged to devel, thanks.

maharmstone:
Thanks both

PR #877, opened 2024-08-21
btrfs-progs: mkfs: add new option --ro-subvol

maharmstone:
Adds an --ro-subvol option to mkfs.btrfs, which allows the creation of read-only subvolumes when using --rootdir.

### DIFF
--- a/Documentation/mkfs.btrfs.rst
+++ b/Documentation/mkfs.btrfs.rst
@@ -163,6 +163,7 @@ OPTIONS
         *flags* is an optional comma-separated list of modifiers. Valid choices are:
 
         * *default*: create as default subvolume (this can only be specified once)
+        * *ro*: create as readonly subvolume
 
 --shrink
         Shrink the filesystem to its minimal size, only works with *--rootdir* option.
--- a/Documentation/mkfs.btrfs.rst
+++ b/Documentation/mkfs.btrfs.rst
@@ -155,10 +155,14 @@ OPTIONS
                 contain the files from *rootdir*. Since version 4.14.1 the filesystem size is
                 not minimized. Please see option *--shrink* if you need that functionality.
 
--u|--subvol <subdir>
+-u|--subvol <subdir>:<flags>
         Specify that *subdir* is to be created as a subvolume rather than a regular
         directory.  The option *--rootdir* must also be specified, and *subdir* must be an
         existing subdirectory within it.  This option can be specified multiple times.
+
+        *flags* is an optional comma-separated list of modifiers. Valid choices are:
+
+        * *default*: create as default subvolume (this can only be specified once)
 
 --shrink
         Shrink the filesystem to its minimal size, only works with *--rootdir* option.
--- a/common/root-tree-utils.c
+++ b/common/root-tree-utils.c
@@ -66,7 +66,8 @@ error:
  * The created tree root would have its root_ref as 1.
  * Thus for subvolumes caller needs to properly add ROOT_BACKREF items.
  */
-int btrfs_make_subvolume(struct btrfs_trans_handle *trans, u64 objectid)
+int btrfs_make_subvolume(struct btrfs_trans_handle *trans, u64 objectid,
+			 bool readonly)
 {
 	struct btrfs_fs_info *fs_info = trans->fs_info;
 	struct btrfs_root *root;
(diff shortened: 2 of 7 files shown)


maharmstone:
This will hopefully be the last of the --subvol patches.

The BTRFS_INODE_ROOT_ITEM_INIT bit is there to prevent btrfs_check_and_init_root_item in the kernel clearing the flags on mount.

DaanDeMeyer:
Hmm, this approach means that the default subvolume specified with --default-subvol can't be made read-only? Maybe it's better to make --default-subvol operate on an existing --subvol or --ro-subvol to avoid option explosion.

maharmstone:
> Hmm, this approach means that the default subvolume specified with --default-subvol can't be made read-only? Maybe it's better to make --default-subvol operate on an existing --subvol or --ro-subvol to avoid option explosion.

Yeah, I sort of assumed that you'd always want the default subvolume to be RW - but I suppose not, in the case of immutable distributions etc. Maybe something like `--subvol dir:ro,default`, with a colon separating the directory name from the flags?

DaanDeMeyer:
Both approaches are fine with me.

adam900710:
If we're goiing the `--subvol dir:ro` way, I guess we can remove the existing `--default-subvol` option.

IMHO the `--subvol dir:ro,default` looks a little more generic.

maharmstone:
I've pushed a new branch implementing the changes as discussed.

DaanDeMeyer:
All working as expected from my testing

yuwata:
Hm, then we cannot create subvolume containing colon??

maharmstone:
> Hm, then we cannot create subvolume containing colon??

I borrowed the syntax from Docker's `-v` for bind mounts. All characters apart from / and \0 are legal in Linux filenames, so there's no safe choice.
Colons have the advantage that they're not legal in filenames on Windows, so people tend to avoid them even elsewhere.

yuwata:
@maharmstone Thank you for the reply. I just thought the change may cause backward compatibility breakage. Bu I noticed that --subvol is new in the next version. Sorry for the noise.

adam900710:
`devel` branch updated to remove the `--default-subvol` option.

Would you please update your first patch just a little? As now there is no `--default-subvol` option to replace.

Otherwise looks good to me.

DaanDeMeyer:
@adam900710 @maharmstone So in systemd-repart we'll keep `DefaultSubvolume=` as a separate setting because there can only be one default subvolume, which maps more cleanly to a separate setting than a flag. So you could consider keeping `--default-subvol` but instead of creating a new subvolume it would make a subvolume specified with `--subvol` the default one. And then if `--default-subvol` is specified multiple times you'd either override the previous value or fail.

Up to you though, I don't mind either API.

maharmstone:
> @maharmstone Thank you for the reply. I just thought the change may cause backward compatibility breakage. Bu I noticed that --subvol is new in the next version. Sorry for the noise.

No worries. I realized after writing that there's an easy solution, which I've included: changing it so it only looks at the last colon. So you do e.g. `--subvol edge:case:` to specify the directory `edge:case` with no flags.

> `devel` branch updated to remove the `--default-subvol` option.
> 
> Would you please update your first patch just a little? As now there is no `--default-subvol` option to replace.
> 
> Otherwise looks good to me.

Thanks, have done

> @adam900710 @maharmstone So in systemd-repart we'll keep `DefaultSubvolume=` as a separate setting because there can only be one default subvolume, which maps more cleanly to a separate setting than a flag. So you could consider keeping `--default-subvol` but instead of creating a new subvolume it would make a subvolume specified with `--subvol` the default one. And then if `--default-subvol` is specified multiple times you'd either override the previous value or fail.
> 
> Up to you though, I don't mind either API.

I've kept it as it is - I think it's best if we don't introduce two ways of doing things, if it's a new flag.

adam900710 (approved):
Looks good to me. If the interface is fully determined, I'll merge the PR soon.

I'll leave for several days just in case if you want to change the interface.

DaanDeMeyer:
@adam900710 We're all happy with this API, so would be great if this can be merged, then the corresponding PR in systemd can be merged as well.

adam900710:
Thanks for the feedback. Just to be clear, although merged into `devel` branch, the feature will only be included in the next major release (v6.11).
So if something went wrong we still have some time to fix.

PR #879, opened 2024-08-21
btrfs-progs: add rudimentary log checking

maharmstone:
Currently the transaction log is more or less ignored by btrfs check, meaning that it's possible for a FS with a corrupt log to pass btrfs check, but be immediately corrupted by the kernel when it's mounted.

This patch adds a check that if there's an inode in the log, any pending writes also have corresponding csum entries.


maharmstone:
Resubmission of log-checking patch

check/main.c
```diff
@@ -9670,6 +9670,272 @@ static int zero_log_tree(struct btrfs_root *root)
...
+		}
+
+next:
+		path.slots[0]++;
+	}
+
+	btrfs_release_path(&path);
+
+	if (ret >= 0)
+		ret = length == 0 ? 1 : 0;
```
adam900710, line 9741: Not sure if we should return 0 if we have not found the item.
Normally we return 0 if everything is fine, the only exception is for functions returning bool (e.g. `check_tree_block_fsid()` from kernel).

check/main.c
```diff
@@ -9670,6 +9670,272 @@ static int zero_log_tree(struct btrfs_root *root)
...
+	check.owner_root = btrfs_root_id(tmp_root);
+	check.transid = btrfs_root_generation(&tmp_root->root_item);
+	check.level = btrfs_root_level(&tmp_root->root_item);
+
+	tmp_root->node = read_tree_block(gfs_info,
+					 btrfs_root_bytenr(&tmp_root->root_item),
+					 &check);
+	if (IS_ERR(tmp_root->node)) {
+		tmp_root->node = NULL;
+		return 1;
```
adam900710, line 9852: You can directly return `PTR_ERR()` instead of 1.

check/main.c
```diff
@@ -9670,6 +9670,272 @@ static int zero_log_tree(struct btrfs_root *root)
...
+	}
+
+	if (btrfs_header_level(tmp_root->node) != btrfs_root_level(&tmp_root->root_item)) {
+		error("root [%llu %llu] level %d does not match %d",
+			tmp_root->root_key.objectid,
+			tmp_root->root_key.offset,
+			btrfs_header_level(tmp_root->node),
+			btrfs_root_level(&tmp_root->root_item));
+		return 1;
```
adam900710, line 9861: The same here, better return some error code (e.g. -EIO) instead of 1.

check/main.c
```diff
@@ -9670,6 +9670,272 @@ static int zero_log_tree(struct btrfs_root *root)
...
+	btrfs_release_path(&path);
+
+	if (ret >= 0)
+		ret = length == 0 ? 1 : 0;
+
+	return ret;
+}
+
+static int check_log_root(struct btrfs_root *root, struct cache_tree *root_cache,
+			  struct walk_control *wc)
```
adam900710, line 9747: The `wc` is not utilized at all.

The walk control structure is mostly to handle tree blocks shared by different subvolumes, but for log tree it's pretty simple and no need to utilize that complex facility.

check/main.c
```diff
@@ -9670,6 +9670,272 @@ static int zero_log_tree(struct btrfs_root *root)
...
+			if (tmp_root.node)
+				free_extent_buffer(tmp_root.node);
+		}
+
+		path.slots[0]++;
+	}
+out:
+	btrfs_release_path(&path);
+	if (err)
+		free_extent_cache_tree(&wc.shared);
```
adam900710, line 9932: If we do not utilize walk control, there is no need to free or cleanup it either.

adam900710 (commented):
Overall looks good to me.

Just some small problems related to return value and an unnecessary parameter.

maharmstone:
Thanks Qu. I've pushed a version with the changes made.

maharmstone:
Thanks Qu. Yes, that's left over from the first draft, it should have been taken out.

check/main.c
```diff
@@ -9670,6 +9670,266 @@ static int zero_log_tree(struct btrfs_root *root)
...
+{
+	struct btrfs_path path = { 0 };
+	struct btrfs_key key;
+	struct extent_buffer *leaf;
+	u16 csum_size = gfs_info->csum_size;
+	u16 num_entries;
+	u64 data_len;
+	int ret;
+
+	key.objectid = BTRFS_EXTENT_CSUM_OBJECTID;
```
kdave, line 9683: You can also define the key inside declaration when it's near the top like that.

check/main.c
```diff
@@ -9670,6 +9670,266 @@ static int zero_log_tree(struct btrfs_root *root)
 	return ret;
 }
 
+static int check_log_csum(struct btrfs_root *root, u64 addr, u64 length)
+{
+	struct btrfs_path path = { 0 };
+	struct btrfs_key key;
+	struct extent_buffer *leaf;
+	u16 csum_size = gfs_info->csum_size;
```
kdave, line 9678: Used only once, so can be dropped.

check/main.c
```diff
@@ -9670,6 +9670,266 @@ static int zero_log_tree(struct btrfs_root *root)
...
+	u16 num_entries;
```
kdave, line 9679: This is an unsafe type used later for multiplication, u32 would be safer.

check/main.c
```diff
@@ -9670,6 +9670,266 @@ static int zero_log_tree(struct btrfs_root *root)
...
+			break;
+
+		if (key.objectid != BTRFS_EXTENT_CSUM_OBJECTID ||
+		    key.type != BTRFS_EXTENT_CSUM_KEY)
+			goto next;
+
+		if (key.offset >= addr + length)
+			break;
+
+		num_entries = btrfs_item_size(leaf, path.slots[0]) / csum_size;
```
kdave, line 9721: item_size returns u32, the type should match num_entries

check/main.c
```diff
@@ -9670,6 +9670,266 @@ static int zero_log_tree(struct btrfs_root *root)
...
+			length = btrfs_file_extent_num_bytes(leaf, fi);
+
+			ret = check_log_csum(root, addr, length);
+			if (ret < 0) {
+				err = 1;
+				break;
+			}
+
+			if (ret) {
+				error("csum missing in log (root %llu, inode %llu, "
```
kdave, line 9808: Please don't split error messages, this should be greppable.

check/main.c
```diff
@@ -9670,6 +9670,266 @@ static int zero_log_tree(struct btrfs_root *root)
...
+
+	btrfs_release_path(&path);
+
+	return err;
+}
+
+static int load_log_root(u64 root_id, struct btrfs_path *path,
+			 struct btrfs_root *tmp_root)
+{
+	struct extent_buffer *l;
```
kdave, line 9828: Please don't use single letter variables for anything else than 'i' in loops. Commonly the extent buffer is either eb or leaf (as it is in another function).

check/main.c
```diff
@@ -10468,9 +10728,21 @@ static int cmd_check(const struct cmd_struct *cmd, int argc, char **argv)
...
+		err |= !!ret;
+	} else {
+		fprintf(stderr,
+		"[1/8] checking log skipped (none written)\n");
+	}
+
 	if (!init_extent_tree) {
 		if (!g_task_ctx.progress_enabled) {
-			fprintf(stderr, "[1/7] checking root items\n");
+			fprintf(stderr, "[2/8] checking root items\n");
```
kdave, line 10745: Suggestion for a future cleanup, count the phases in a variable so we don't have to update all the messages each time we decide to add new or change order.

check/main.c
maharmstone: My mistake, I thought the max size of an item was 65535.

PR #882, opened 2024-08-30
btrfs-progs: mkfs: add --compression

maharmstone:
Adds a --compression option to mkfs.btrfs when using --rootdir.

### DIFF
--- a/Documentation/mkfs.btrfs.rst
+++ b/Documentation/mkfs.btrfs.rst
@@ -214,8 +214,8 @@ OPTIONS
 
 --compress <algo>:<level>
         Try to compress files when using *--rootdir*.  Supported values for *algo* are
-        *no* (the default) and *zlib*.  The optional value *level* is a
-        compression level, from 1 to 9 for ZLIB.
+        *no* (the default), *zlib*, and *zstd*.  The optional value *level* is a
+        compression level, from 1 to 9 for ZLIB and from 1 to 15 for ZSTD.
 
         As with the kernel, :command:`mkfs.btrfs` won't write compressed extents when
         they would be larger than the uncompressed versions, and will mark a file as
--- a/Documentation/mkfs.btrfs.rst
+++ b/Documentation/mkfs.btrfs.rst
@@ -214,7 +214,7 @@ OPTIONS
 
 --compress <algo>:<level>
         Try to compress files when using *--rootdir*.  Supported values for *algo* are
-        *no* (the default), *zlib*, and *zstd*.  The optional value *level* is a
+        *no* (the default), *zlib*, *lzo*, and *zstd*.  The optional value *level* is a
         compression level, from 1 to 9 for ZLIB and from 1 to 15 for ZSTD.
 
         As with the kernel, :command:`mkfs.btrfs` won't write compressed extents when
--- a/Documentation/mkfs.btrfs.rst
+++ b/Documentation/mkfs.btrfs.rst
@@ -212,6 +212,15 @@ OPTIONS
 
                 $ mkfs.btrfs -O list-all
 
+--compress <algo>:<level>
+        Try to compress files when using *--rootdir*.  Supported values for *algo* are
+        *no* (the default) and *zlib*.  The optional value *level* is a
+        compression level, from 1 to 9 for ZLIB.
+
+        As with the kernel, :command:`mkfs.btrfs` won't write compressed extents when
+        they would be larger than the uncompressed versions, and will mark a file as
+        `nocompress` if its beginning is found to be incompressible.
+
 -f|--force
         Forcibly overwrite the block devices when an existing filesystem is detected.
         By default, :command:`mkfs.btrfs` will utilize *libblkid* to check for any known
--- a/common/extent-tree-utils.c
+++ b/common/extent-tree-utils.c
@@ -285,3 +285,93 @@ int btrfs_record_file_extent(struct btrfs_trans_handle *trans,
 	}
 	return ret;
 }
+
+int btrfs_record_file_extent_comp(struct btrfs_trans_handle *trans,
+				  struct btrfs_root *root, u64 objectid,
+				  struct btrfs_inode_item *inode,
+				  u64 file_pos, u64 disk_bytenr,
+				  u64 extent_num_bytes, u64 ram_bytes,
+				  enum btrfs_compression_type comp)
+{
+	struct btrfs_fs_info *info = root->fs_info;
+	struct btrfs_root *extent_root = btrfs_extent_root(info, disk_bytenr);
+	struct extent_buffer *leaf;
+	struct btrfs_file_extent_item *fi;
+	struct btrfs_key ins_key;
+	struct btrfs_path *path;
+	struct btrfs_extent_item *ei;
+	u64 nbytes;
+	u64 extent_bytenr;
+	u64 extent_offset;
+	int ret = 0;
+
+	path = btrfs_alloc_path();
+	if (!path)
+		return -ENOMEM;
+
+	ins_key.objectid = disk_bytenr;
+	ins_key.type = BTRFS_EXTENT_ITEM_KEY;
+	ins_key.offset = extent_num_bytes;
+
+	ret = btrfs_insert_empty_item(trans, extent_root, path,
+					&ins_key, sizeof(*ei));
+	if (ret == 0) {
+		leaf = path->nodes[0];
+		ei = btrfs_item_ptr(leaf, path->slots[0],
+					struct btrfs_extent_item);
+
+		btrfs_set_extent_refs(leaf, ei, 0);
+		btrfs_set_extent_generation(leaf, ei, trans->transid);
+		btrfs_set_extent_flags(leaf, ei,
+					BTRFS_EXTENT_FLAG_DATA);
+		btrfs_mark_buffer_dirty(leaf);
+
+		ret = btrfs_update_block_group(trans, disk_bytenr,
+					       extent_num_bytes, 1, 0);
+		if (ret)
+			goto fail;
+	} else if (ret != -EEXIST) {
+		goto fail;
+	}
+
+	ret = remove_from_free_space_tree(trans, disk_bytenr, extent_num_bytes);
+	if (ret)
+		goto fail;
+
+	btrfs_run_delayed_refs(trans, -1);
+	extent_bytenr = disk_bytenr;
+	extent_offset = 0;
+
+	btrfs_release_path(path);
+	ins_key.objectid = objectid;
+	ins_key.type = BTRFS_EXTENT_DATA_KEY;
+	ins_key.offset = file_pos;
+	ret = btrfs_insert_empty_item(trans, root, path, &ins_key, sizeof(*fi));
+	if (ret)
+		goto fail;
+	leaf = path->nodes[0];
+	fi = btrfs_item_ptr(leaf, path->slots[0], struct btrfs_file_extent_item);
+	btrfs_set_file_extent_generation(leaf, fi, trans->transid);
+	btrfs_set_file_extent_type(leaf, fi, BTRFS_FILE_EXTENT_REG);
+	btrfs_set_file_extent_disk_bytenr(leaf, fi, extent_bytenr);
+	btrfs_set_file_extent_disk_num_bytes(leaf, fi, extent_num_bytes);
+	btrfs_set_file_extent_offset(leaf, fi, extent_offset);
+	btrfs_set_file_extent_num_bytes(leaf, fi, ram_bytes);
+	btrfs_set_file_extent_ram_bytes(leaf, fi, ram_bytes);
+	btrfs_set_file_extent_compression(leaf, fi, comp);
+	btrfs_set_file_extent_encryption(leaf, fi, 0);
+	btrfs_set_file_extent_other_encoding(leaf, fi, 0);
+	btrfs_mark_buffer_dirty(leaf);
+
+	nbytes = btrfs_stack_inode_nbytes(inode) + ram_bytes;
+	btrfs_set_stack_inode_nbytes(inode, nbytes);
+	btrfs_release_path(path);
+
+	ret = btrfs_inc_extent_ref(trans, extent_bytenr, extent_num_bytes,
+				   0, root->root_key.objectid, objectid,
+				   file_pos - extent_offset);
+
+fail:
+	btrfs_free_path(path);
+	return ret;
+}
--- a/convert/source-ext2.c
+++ b/convert/source-ext2.c
@@ -457,7 +457,8 @@ static int ext2_create_file_extents(struct btrfs_trans_handle *trans,
 		if (num_bytes > inode_size)
 			num_bytes = inode_size;
 		ret = btrfs_insert_inline_extent(trans, root, objectid,
-						 0, buffer, num_bytes);
+						 0, buffer, num_bytes,
+						 0, num_bytes);
 		if (ret)
 			goto fail;
 		nbytes = btrfs_stack_inode_nbytes(btrfs_inode) + num_bytes;
@@ -506,7 +507,8 @@ static int ext2_create_symlink(struct btrfs_trans_handle *trans,
 	pathname = (char *)&(ext2_inode->i_block[0]);
 	BUG_ON(pathname[inode_size] != 0);
 	ret = btrfs_insert_inline_extent(trans, root, objectid, 0,
-					 pathname, inode_size);
+					 pathname, inode_size,
+					 0, inode_size);
 	btrfs_set_stack_inode_nbytes(btrfs_inode, inode_size);
 	return ret;
 }
--- a/convert/source-reiserfs.c
+++ b/convert/source-reiserfs.c
@@ -384,7 +384,8 @@ static int reiserfs_convert_tail(struct btrfs_trans_handle *trans,
 				      length, offset, convert_flags);
 
 	ret = btrfs_insert_inline_extent(trans, root, objectid,
-					 offset, body, length);
+					 offset, body, length,
+					 0, length);
 	if (ret)
 		return ret;
 
@@ -544,7 +545,7 @@ static int reiserfs_copy_symlink(struct btrfs_trans_handle *trans,
 		goto fail;
 	}
 	ret = btrfs_insert_inline_extent(trans, root, objectid, 0,
-					 symlink, len);
+					 symlink, len, 0, len);
 	btrfs_set_stack_inode_nbytes(btrfs_inode, len);
 fail:
 	pathrelse(&path);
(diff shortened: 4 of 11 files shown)


maharmstone:
Thanks Qu

mkfs/main.c
```diff
@@ -443,6 +443,7 @@ static const char * const mkfs_usage[] = {
 	OPTLINE("-u|--subvol SUBDIR", "create SUBDIR as subvolume rather than normal directory, can be specified multiple times"),
 	OPTLINE("--shrink", "(with --rootdir) shrink the filled filesystem to minimal size"),
 	OPTLINE("-K|--nodiscard", "do not perform whole device TRIM"),
+	OPTLINE("-z|--compression ALGO", "compression algorithm to use, none (default), zstd"),
```
adam900710, line 446: Although it's a nobrainer to choose zstd, mind to add other algos too?

And I'm not sure if David is fine with short options, normally unless it's a very common feature, we only go long options.

mkfs/rootdir.c
```diff
@@ -286,33 +397,108 @@ static int write_extent(struct btrfs_trans_handle *trans,
...
+					   bytes_read, comp_buf);
+		if (comp_ret < 0)
+			return comp_ret;
+		else if (comp_ret == 0)
+			do_comp = false;
+
+		/* If we've just marked this inode nocompress, increase the
+		 * buffer size and read the rest of the extent.  */
+		if (btrfs_stack_inode_flags(btrfs_inode) & BTRFS_INODE_NOCOMPRESS) {
+			buf_size = SZ_1M;
```
adam900710, line 415: Can we just use 128K buf_size for all?

I think it's already a large enough buffer size. This buffer size change can be done in the previous commit.

mkfs/rootdir.c
```diff
@@ -286,33 +397,108 @@ static int write_extent(struct btrfs_trans_handle *trans,
...
+				bytes_read += ret_read;
+			}
+		}
+	}
 
-	ret = btrfs_reserve_extent(trans, root, to_read + padding, 0, 0,
+	if (do_comp) {
+		u64 flags, features;
+
+		to_write = DIV_ROUND_UP(comp_ret, sectorsize) * sectorsize;
```
adam900710, line 450: What about plain `round_up(comp_ret)`?

mkfs/rootdir.c
```diff
@@ -286,33 +397,108 @@ static int write_extent(struct btrfs_trans_handle *trans,
...
+		write_buf = comp_buf;
+		memset(write_buf + comp_ret, 0, to_write - comp_ret);
```
adam900710, line 452: I think this patter can also applied to the previous patch to do the padding, without introducing a dedicated `padding` variable.

mkfs/rootdir.c
adam900710: Nevermind 1MiB is much preferable to reduce fragmentation, so it's still better to use larger buffer size when possible.

My bad, it's not about read performance but fragmentation.

convert/source-ext2.c
```diff
@@ -445,7 +445,8 @@ static int ext2_create_file_extents(struct btrfs_trans_handle *trans,
 		if (num_bytes > inode_size)
 			num_bytes = inode_size;
 		ret = btrfs_insert_inline_extent(trans, root, objectid,
-						 0, buffer, num_bytes);
+						 0, buffer, num_bytes,
+						 0, inode_size + 1);
```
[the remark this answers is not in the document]
maharmstone: Oops, this one's a mistake - it should be `num_bytes` again. Thanks

convert/source-ext2.c
```diff
@@ -489,7 +490,8 @@ static int ext2_create_symlink(struct btrfs_trans_handle *trans,
 	pathname = (char *)&(ext2_inode->i_block[0]);
 	BUG_ON(pathname[inode_size] != 0);
 	ret = btrfs_insert_inline_extent(trans, root, objectid, 0,
-					 pathname, inode_size + 1);
+					 pathname, inode_size + 1,
+					 0, inode_size + 1);
```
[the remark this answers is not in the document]
maharmstone: `inode_size + 1` is here because I've added a separate `ram_bytes` param to `btrfs_insert_inline_extent`, because it's no longer necessarily the same as `size`.

convert/source-reiserfs.c
```diff
@@ -538,7 +539,7 @@ static int reiserfs_copy_symlink(struct btrfs_trans_handle *trans,
 	len = get_ih_item_len(tp_item_head(&path));
 
 	ret = btrfs_insert_inline_extent(trans, root, objectid, 0,
-					 symlink, len + 1);
+					 symlink, len + 1, 0, len + 1);
```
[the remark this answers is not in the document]
maharmstone: As in `ext2_create_symlink`

mkfs/main.c
maharmstone: Yes, I'm planning to add zlib and lzo too.

I can remove `-z` if that's what you want, but it's a very common option to mean "compression" (e.g. tar, rsync)

mkfs/rootdir.c
maharmstone: Thanks, I wasn't aware of this!

mkfs/rootdir.c
```diff
@@ -241,6 +241,87 @@ static int add_symbolic_link(struct btrfs_trans_handle *trans,
 	return ret;
 }
 
+static int write_extent(struct btrfs_trans_handle *trans,
+			struct btrfs_root *root,
+			struct btrfs_inode_item *btrfs_inode, u64 objectid,
+			int fd, u64 *file_pos, char *buf, u64 *left,
```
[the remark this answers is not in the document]
maharmstone: `left` is there as a cap on the buffer size, but if we know the file position it'd be simpler to replace it with the inode size.

maharmstone:
Pushed new version with changes discussed.

mkfs/main.c
adam900710: Considering the option only makes sense for `--rootdir`, I'm not too confident about a short option.

And we can always add new short option if it's common utilized later, so far the long option only way looks safer to merge.

convert/source-ext2.c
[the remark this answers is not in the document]
adam900710: In fact this is a minor problem in the existing convert code. The resulted symlink would be one byte larger than the original path (but inode size is still correct).

E.g.
The original ext4 has a link to file name `path`:
```
lrwxrwxrwx 1 root root 4 Sep  3 10:27 path -> path
```

But after convert, the result looks like this:
```
	item 10 key (267 INODE_ITEM 0) itemoff 15543 itemsize 160
		generation 1 transid 1 size 4 nbytes 5
		block group 0 mode 120777 links 1 uid 0 gid 0 rdev 0
		sequence 0 flags 0x0(none)
		atime 1725325025.930020384 (2024-09-03 10:27:05)
		ctime 1725325023.110085974 (2024-09-03 10:27:03)
		mtime 1725325023.110085974 (2024-09-03 10:27:03)
		otime 1725325023.110085974 (2024-09-03 10:27:03)
	item 11 key (267 INODE_REF 256) itemoff 15529 itemsize 14
		index 3 namelen 4 name: path
	item 12 key (267 EXTENT_DATA 0) itemoff 15503 itemsize 26
		generation 4 type 0 (inline)
		inline extent data size 5 ram_bytes 5 compression 0 (none)
```
Note the link has one extra byte.

It's definitely not your fault nor it's causing problems, but it will also help a lot if you want to fix it.

convert/source-reiserfs.c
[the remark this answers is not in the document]
adam900710: The same bug as `ext2_create_symlink`

mkfs/rootdir.c
```diff
@@ -241,6 +241,79 @@ static int add_symbolic_link(struct btrfs_trans_handle *trans,
...
+{
+	int ret;
+	u32 sectorsize = root->fs_info->sectorsize;
+	u64 bytes_read, first_block, to_read, to_write;
+	struct btrfs_key key;
+
+	if (file_pos + MAX_EXTENT_SIZE > size)
+		to_read = size - file_pos;
+	else
+		to_read = MAX_EXTENT_SIZE;
```
adam900710: What about `to_read = min(file_pos + MAX_EXTENT_SIZE, size) - file_pos;`?

adam900710 (approved):
Just a really small nitpick. Otherwise looks perfect to me.

convert/source-ext2.c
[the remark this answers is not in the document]
adam900710: And the fix for that is here: https://github.com/kdave/btrfs-progs/pull/884

DaanDeMeyer:
@adam900710 Any chance this could be considered for 6.11 as well? Then we can land support for this in systemd-repart in the upcoming systemd 257 release.

convert/source-ext2.c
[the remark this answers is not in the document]
adam900710: The proper fix for that bug is here: https://github.com/kdave/btrfs-progs/pull/885

In fact the +1 is already causing `btrfs/012` random failures.
So that has to be fixed anyway.

adam900710:
David is not hurry to push so many features in one single release.

And considering this is only for compression, it shouldn't cause too much change except the image size.

DaanDeMeyer:
> David is not hurry to push so many features in one single release.
> 
> And considering this is only for compression, it shouldn't cause too much change except the image size.

The image size is roughly halved by enabling this in my case, which can make a huge difference.

adam900710:
Compression is indeed awesome to reduce the image size, but here we really want to go a more stable methods towards new features.

Personally speaking I still want more review on the final patch adding the compression support.

@kdave may give a better roadmap on this, and do the final call.

mkfs/rootdir.c
maharmstone: No problem, makes sense

maharmstone:
Pushed with latest change

convert/source-ext2.c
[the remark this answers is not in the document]
maharmstone: Okay, I'll rebase once you've committed your fixes.

Conan-Kudo:
It would be really great for us image build tools if this landed in 6.11 / systemd 257. It becomes possible to create Fedora images mostly offline and unprivileged with these features.

Documentation/mkfs.btrfs.rst
```diff
@@ -188,6 +188,12 @@ OPTIONS
 
                 $ mkfs.btrfs -O list-all
 
+--compression <algo>
```
kdave, line 191: This should support all algorithms, lzo and zlib can be added later. Also the compression level should be supported, for zstd it may make sense to use the highest level (matching kernel levels) when creating the fs.

kdave (changes requested):

maharmstone:
I've pushed a new version, which includes setting the level and renaming `--compression` to `--compress`. I've changed it so that the base patch includes zlib and the subsequent patch adds zstd, on the grounds that zlib is a hard dependency but zstd isn't.

maharmstone:
Rebased against devel

maharmstone:
Pushed rebase

mkfs/rootdir.c
```diff
@@ -362,6 +362,76 @@ static int add_symbolic_link(struct btrfs_trans_handle *trans,
...
+	to_read = min(file_pos + MAX_EXTENT_SIZE, size) - file_pos;
+
+	bytes_read = 0;
+
+	while (bytes_read < to_read) {
+		ssize_t ret_read;
+
+		ret_read = pread(fd, buf + bytes_read, to_read - bytes_read,
+				 file_pos + bytes_read);
+		if (ret_read == -1) {
```
adam900710, line 658: Please use `ret_read < 0`, as that's the common practice for detecting an error, not only for the common IO but also all the other btrfs specific calls.

mkfs/rootdir.c
```diff
@@ -362,6 +362,76 @@ static int add_symbolic_link(struct btrfs_trans_handle *trans,
 	return ret;
 }
 
+/*
+ * keep our extent size at 1MB max, this makes it easier to work
+ * inside the tiny block groups created during mkfs
+ */
+#define MAX_EXTENT_SIZE SZ_1M
+
+static int read_and_write_extent(struct btrfs_trans_handle *trans,
```
adam900710, line 629: I found the function name `read_and_write_extent()` a little hard to grasp.

The caller is `add_file_items()`, what about just a simple `add_one_file_item()`.

mkfs/rootdir.c
```diff
@@ -362,6 +362,76 @@ static int add_symbolic_link(struct btrfs_trans_handle *trans,
...
+				 struct btrfs_root *root,
+				 struct btrfs_inode_item *btrfs_inode,
+				 u64 objectid, int fd, u64 file_pos, char *buf,
```
adam900710, line 632: Everything from `fd`, `buf`, `size` (from st structure), and `path_name` is only here to help read the file.
The only thing changing is `file_pos` which is worth a dedicated parameter.

During my initial failed merge, I added an extra helper structure, source_descriptor, to include `fd`, `st`, `path` and `buf`. Later expands the structure to add `compressed_buf` for compression.

Such structure is more expandable than adding more and more parameters to the already long list.

mkfs/rootdir.c
```diff
@@ -403,10 +463,10 @@ static int add_file_items(struct btrfs_trans_handle *trans,
 			goto end;
 		}
 
-		ret_read = pread(fd, buffer, st->st_size, bytes_read);
+		ret_read = pread(fd, buffer, st->st_size, 0);
 		if (ret_read == -1) {
-			error("cannot read %s at offset %llu length %llu: %m",
-				path_name, bytes_read, (unsigned long long)st->st_size);
+			error("cannot read %s at offset %u length %llu: %m",
+				path_name, 0, (unsigned long long)st->st_size);
```
adam900710, line 1010: Since you are here, you can also fix the length output, without using a forced `(unsigned long long)` conversion, but `%zu`.

Documentation/mkfs.btrfs.rst
```diff
@@ -212,6 +212,15 @@ OPTIONS
 
                 $ mkfs.btrfs -O list-all
 
+--compress <algo>:<level>
```
adam900710, line 215: Since `level` is optional, please use `--compress <algo>[:<level>]` instead.

common/extent-tree-utils.c
```diff
@@ -285,3 +285,93 @@ int btrfs_record_file_extent(struct btrfs_trans_handle *trans,
 	}
 	return ret;
 }
+
+int btrfs_record_file_extent_comp(struct btrfs_trans_handle *trans,
```
adam900710, line 289: Please do no use `comp` as a short name for `compress`, it's a little confusing.

But it will no longer be a problem if we address the next comment.

common/extent-tree-utils.c
```diff
@@ -285,3 +285,93 @@ int btrfs_record_file_extent(struct btrfs_trans_handle *trans,
...
+				  struct btrfs_root *root, u64 objectid,
+				  struct btrfs_inode_item *inode,
+				  u64 file_pos, u64 disk_bytenr,
+				  u64 extent_num_bytes, u64 ram_bytes,
+				  enum btrfs_compression_type comp)
```
adam900710, line 294: Instead of the longer and longer parameter list, what about something just like `btrfs_ordered_extent` inside the kernel, to record all the needed info?

So that we do not need two different `btrfs_record_file_extent()`, but a single one to handle them all, just like kernel.

This will need a dedicated cleanup patch first, to introduce some structure like `btrfs_ordered_extent`.

mkfs/main.c
```diff
@@ -443,6 +443,7 @@ static const char * const mkfs_usage[] = {
 	OPTLINE("-u|--subvol TYPE:SUBDIR", "create SUBDIR as subvolume rather than normal directory, can be specified multiple times"),
 	OPTLINE("--shrink", "(with --rootdir) shrink the filled filesystem to minimal size"),
 	OPTLINE("-K|--nodiscard", "do not perform whole device TRIM"),
+	OPTLINE("--compress ALGO:LEVEL", "compression algorithm and level to use; ALGO can be no (default), zlib"),
```
adam900710: Also `[]` for the level.

mkfs/rootdir.c
```diff
@@ -362,6 +369,75 @@ static int add_symbolic_link(struct btrfs_trans_handle *trans,
...
+		if (ret != Z_OK) {
+			error("deflate failed: %s", strm.msg);
+			return -EINVAL;
+		}
+
+		if (strm.avail_out < BTRFS_MAX_COMPRESSED - sectorsize) {
+			u64 flags;
+
+			flags = btrfs_stack_inode_flags(btrfs_inode);
+			flags |= BTRFS_INODE_NOCOMPRESS;
```
adam900710, line 419: I prefer not to pass a btrfs inode and handle its flags inside a compression function.

Such compression function should only do the compression. And let the caller handling the btrfs specific things.
(E.g. if the compressed size is larger than the input,  return `-E2BIG` to inform the caller just like what kernel does).

mkfs/rootdir.c
```diff
@@ -471,8 +672,29 @@ static int add_file_items(struct btrfs_trans_handle *trans,
...
+							  &comp_buf, &comp_size);
+			if (ret < 0)
+				goto end;
+			break;
+		default:
+			ret = 0;
+			break;
+		}
+
+		if (ret == 1) {
```
adam900710, line 1042: Please do not use immediate number other than 0, so this should be `if (ret > 0) {`

adam900710 (commented):
Sorry I merged the branch by incident, and I can not find a way to re-open the merge request.

It turns out that when modifying the code inside my editor, it's much easier to expose problems, than viewing it through github web.
(Not to mention github hides large chunks of code by default).

common/extent-tree-utils.c
adam900710: It turns out that we should not use `btrfs_record_file_extent()` except for convert.

The extra handling inside `btrfs_record_file_extent()` is specially designed for convert, where we may have an existing data extent from the convert image, thus it is not a good way to support compressed data extents.

For mkfs cases, I'll craft a dedicated help for it with a more expandable interface during my cleanup.

adam900710:
BTW, I have submitted the series cleaning up the file extent insertion code:
https://github.com/kdave/btrfs-progs/pull/919

Now for mkfs rootdir usage, we should go the new helper, `insert_reserved_file_extent()`, and it accepts an on-stack file extent item pointer, so caller can specify all the members, just like what we did inside the kernel.

This should allow the function to be utilized by both regular and compressed file extents, again, just like the kernel.

mkfs/rootdir.c
maharmstone: Because it's not adding a file item, it's adding an extent.

I'll rename it to add_file_item_extent.

PR #885, opened 2024-09-03
btrfs-progs: convert: fix the invalid regular extents for symbol links

adam900710:
Test case btrfs/012 fails randomly after the rework to use fsstress to
populate the fs.

It turns out that, if fsstress creates a symbol link whose target is
4095 bytes (at the max size limit), btrfs-convert will create a regular
extent instead of the expected inline one.

This regular extent for symbol link inodes will be rejected by kernel,
resulting test case failure.

The reason that btrfs-convert created such regular extent is,
btrfs-convert accidentally added one byte for the terminating NUL,
enlarge the should-be inlined file extent to be a regular one.

The first patch will fix the bug.
Then two patches to enahnce btrfs-check to detect the error (regular and
lowmem mode)
Eventually a dedicated test case for btrfs-convert, so in the future we
won't cause the problem again.

### DIFF
--- a/check/main.c
+++ b/check/main.c
@@ -1745,6 +1745,13 @@ static int process_file_extent(struct btrfs_root *root,
 			rec->errors |= I_ERR_BAD_FILE_EXTENT;
 		if (disk_bytenr > 0)
 			rec->found_size += num_bytes;
+		/*
+		 * Symbolic links should only have inlined extents.
+		 * A regular extent means it's already too large to
+		 * be inlined.
+		 */
+		if (S_ISLNK(rec->imode))
+			rec->errors |= I_ERR_FILE_EXTENT_TOO_LARGE;
 	} else {
 		rec->errors |= I_ERR_BAD_FILE_EXTENT;
 	}
--- a/check/mode-lowmem.c
+++ b/check/mode-lowmem.c
@@ -3351,6 +3351,31 @@ out_no_release:
 	return err;
 }
 
+static int read_inode_item(struct btrfs_root *root,
+			   u64 ino, struct btrfs_inode_item *ret_ii)
+{
+	struct btrfs_path path = { 0 };
+	struct btrfs_key key = {
+		.objectid = ino,
+		.type = BTRFS_INODE_ITEM_KEY,
+		.offset = 0
+	};
+	int ret;
+
+	ret = btrfs_search_slot(NULL, root, &key, &path, 0, 0);
+	if (ret > 0)
+		ret = -ENOENT;
+	if (ret < 0)
+		goto out;
+
+	read_extent_buffer(path.nodes[0], ret_ii,
+			   btrfs_item_ptr_offset(path.nodes[0], path.slots[0]),
+			   sizeof(*ret_ii));
+out:
+	btrfs_release_path(&path);
+	return ret;
+}
+
 /*
  * Check EXTENT_DATA item, mainly for its dbackref in extent tree
  *
@@ -3371,6 +3396,7 @@ static int check_extent_data_item(struct btrfs_root *root,
 	struct btrfs_extent_item *ei;
 	struct btrfs_extent_inline_ref *iref;
 	struct btrfs_extent_data_ref *dref;
+	struct btrfs_inode_item inode_item;
 	u64 owner;
 	u64 disk_bytenr;
 	u64 disk_num_bytes;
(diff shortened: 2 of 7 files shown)


maharmstone:
From what I gather, the maximum size of a symlink is PATH_MAX, which is 4096 bytes (see `getname_flags` in fs/namei.c). I think quite possibly the references to the sector size in your patches should actually be PATH_MAX.

check/main.c
```diff
@@ -1745,6 +1745,13 @@ static int process_file_extent(struct btrfs_root *root,
 			rec->errors |= I_ERR_BAD_FILE_EXTENT;
 		if (disk_bytenr > 0)
 			rec->found_size += num_bytes;
+		/*
+		 * Symbol link should only have inlined extent.
```
kdave, line 1749: Symbol -> Symbolic

check/mode-lowmem.c
```diff
@@ -3400,6 +3426,24 @@ static int check_extent_data_item(struct btrfs_root *root,
 	extent_num_bytes = btrfs_file_extent_num_bytes(eb, fi);
 	offset = btrfs_file_extent_offset(eb, fi);
 
+	/*
+	 * There is a regular/preallocated data extent. Make sure the owning
+	 * inode is not a symbol link.
```
kdave, line 3431: symbol -> symbolic

check/mode-lowmem.c
```diff
@@ -3351,6 +3351,31 @@ static int repair_extent_data_item(struct btrfs_root *root,
 	return err;
 }
 
+static int grab_inode_item(struct btrfs_root *root,
```
kdave, line 3354: The name `grab` is usually used for something regarding references when reading an object, please rename it to `read_inode_item`.

check/mode-lowmem.c
```diff
@@ -3400,6 +3426,24 @@ static int check_extent_data_item(struct btrfs_root *root,
...
+	 */
+	ret = grab_inode_item(root, fi_key.objectid, &inode_item);
+	if (ret < 0) {
+		errno = -ret;
+		error("failed to grab the inode item for inode %llu: %m",
+		      fi_key.objectid);
+		err |= INODE_ITEM_MISSING;
+	}
+	if (S_ISLNK(inode_item.mode)) {
+		error("symbol link at root %lld ino %llu has regular/preallocated extents",
```
kdave, line 3442: symbolic

tests/convert-tests/027-large-symbol-link/test.sh
```diff
@@ -0,0 +1,33 @@
...
+prepare_test_dev 1G
+check_global_prereq mkfs.ext4
+
+link_target=""
+
+# Soft link in btrfs can only have inlined data extent.
+# So here we create a symbol link whose target is 4095 bytes length.
+for ((i = 0; i < 4095; i++)); do
+	link_target+="b"
+done
```
kdave, line 17: You can do it in one go `printf "%0.sb" {1..4095}"`, this is already done in some other tests like mkfs-tests/014-rootdir-inline-extent/test.sh . This would be good to have in a helper so we don't have to figure it out each time, but that's for another patch.

kdave (changes requested):

adam900710:
> From what I gather, the maximum size of a symlink is PATH_MAX, which is 4096 bytes (see `getname_flags` in fs/namei.c). I think quite possibly the references to the sector size in your patches should actually be PATH_MAX.

Indeed, 4095 is a btrfs' internal limits (due to inline extent size). And it can even be much smaller (e.g. 4K nodesize will only have less than 4000 bytes for that).

So I'll rework the wording to make it more explicit that, this is a btrfs specific limit (at least for now).

Considering the kernel has such limits for a long long time, I'm not sure if we can even remove the limits and make it properly to accept 4096 sized soft link.

adam900710:
> > From what I gather, the maximum size of a symlink is PATH_MAX, which is 4096 bytes (see `getname_flags` in fs/namei.c). I think quite possibly the references to the sector size in your patches should actually be PATH_MAX.
> 
> Indeed, 4095 is a btrfs' internal limits (due to inline extent size). And it can even be much smaller (e.g. 4K nodesize will only have less than 4000 bytes for that).
> 
> So I'll rework the wording to make it more explicit that, this is a btrfs specific limit (at least for now).
> 
> Considering the kernel has such limits for a long long time, I'm not sure if we can even remove the limits and make it properly to accept 4096 sized soft link.

In fact, this fix is a little too strict, kernel itself even allows sector sized inline extent. So I should change the code to follow the kernel behavior.
The only limit on the soft link size is inline extent size, not the sector size. This is a little different from our data inline extent.

So I'll need to loose the limit and make sure we can still create an inline extent even at its max size (4096).

adam900710:
> > From what I gather, the maximum size of a symlink is PATH_MAX, which is 4096 bytes (see `getname_flags` in fs/namei.c). I think quite possibly the references to the sector size in your patches should actually be PATH_MAX.
> 
> OK, you're right, I though the PATH_MAX is a limit that's inclusive, but in fact it's a limit including the terminating NUL. And it's inside btrfs (and other fses) that we do not store the terminating NUL.
> 
> I'll reword the commit messages. Thanks a lot for pointing to the proper kernel source file.

And the confusion between sector size and PATH_MAX is, we have two different limits.

For inlined extent data extents, the real limits is `sectorsize - 1`, this means we can create much larger inline extents if the sectorsize is larger (e.g. on 64K page sized system, we can easily create 16K or even larger inlined extents).

For symbolic links, we have extra limits, from the VFS layer, that the real limits is `PATH_MAX - 1`.

And another limits come from the node size `BTRFS_MAX_INLINE_DATA_SIZE()`.

So to properly describe the limits, in the latest version I split the limits into two functions, `btrfs_symlink_max_size()` and `btrfs_data_inline_max_size()` to make the complex situation a little easier to understand.

kdave (approved):
Thanks.

PR #886, opened 2024-09-04
btrfs-progs: add option for recursive subvol snapshots

maharmstone:
Adds an option -R to btrfs subvolume snapshot, corresponding to the flag BTRFS_UTIL_CREATE_SNAPSHOT_RECURSIVE.

This is another resubmission of a missed patch of Omar's from 2018: https://lore.kernel.org/all/e42cdc5d5287269faf4d09e8c9786d0b3adeb658.1516991902.git.osandov@fb.com/


Documentation/btrfs-subvolume.rst
```diff
@@ -252,7 +252,7 @@ show [options] <path>
         -u|--uuid UUID
                 show details about subvolume with the given *UUID*, looked up in *path*
 
-snapshot [-r] [-i <qgroupid>] <source> <dest>|[<dest>/]<name>
+snapshot [-r|-R] [-i <qgroupid>] <source> <dest>|[<dest>/]<name>
```
kdave, line 255: Please also add a long option, in this case it follows `ls` so -R and --recursive.

Documentation/btrfs-subvolume.rst
```diff
@@ -263,6 +263,9 @@ snapshot [-r] [-i <qgroupid>] <source> <dest>|[<dest>/]<name>
 
         -r
                 Make the new snapshot read only.
+        -R
+                Recursively snapshot subvolumes beneath the source. This option cannot be
```
kdave, line 267: This sounds like limitation but we can't do better in user space. This is possible in kernel but is quite complicated so we must exclude -r and -R for now. As a suggestion to documenation, the subvolume can be mad read-only afterwards by changing the property to `ro`.

And as this is using libbtrfsutil the same limitations apply (and need to be documented): it's not atomic and depends on the capabilities to enumerate the subvolumes. Similar to what the recursive deletion has, we might need a special section in documentation for that so we can refer to it  instead of repeating the whole text.

kdave (changes requested):

Documentation/btrfs-subvolume.rst
maharmstone: Thanks, I've added a bit to the man page.

I might also look into fixing the kernel so that this can be done properly.

Documentation/btrfs-subvolume.rst
maharmstone: Done

maharmstone:
Pushed new version

Documentation/btrfs-subvolume.rst
adam900710: I guess David also means the long option should also show up in the doc like:

`snapshjot [-r|-R|--recursive] [-i <qgroupid>] <source> <dest>|[<dest>/]<name`

maharmstone: Oh, okay, have done. I was following what other commands do in this case, but it looks like it's inconsistent.

Documentation/btrfs-subvolume.rst
```diff
@@ -252,17 +252,23 @@ show [options] <path>
...
-snapshot [-r] [-i <qgroupid>] <source> <dest>|[<dest>/]<name>
+snapshot [-r|-R|--recursive] [-i <qgroupid>] <source> <dest>|[<dest>/]<name>
         Create a snapshot of the subvolume *source* with the
         name *name* in the *dest* directory.
 
         If only *dest* is given, the subvolume will be named the basename of *source*.
         If *source* is not a subvolume, btrfs returns an error.
 
+        If you wish to recursively create a readonly snapshot, you can run
+        :command:`btrfs property set <path> ro true` on each subvolume after this command completes.
```
kdave, line 263: For clarity I'd suggest to mention that it has to be done from the leaves up in the tree consisting of subvolumes. Which may not be that easy to find out manually, so this could be a separate command for `subvolume` (in a separate patch). For parity an option to the snapshot command can be added that should read like it's doing the read-only change after the snapshots. This can be also added separatelly, we need to define the use cases first.

Documentation/btrfs-subvolume.rst
kdave: This still does not mention the limitations, nor referes to a section where it would be explained.

kdave (changes requested):

adam900710:
The PR is pretty old, no longer apply.

If you still want to implement this feature, please open a new PR.

PR #891, opened 2024-09-12
btrfs-progs: check: report deprecated inode cache as errors

adam900710:
There are two reports that inode cache is causing kernel to reject those
data backrefs.

It turns out the long to be deprecated inode cache is biting us
unexpectedly.
(And the design of those special inodes has no intention to properly
fill its members like mode and transid correctly).

However the original mode btrfs check is not updated to detect them, in
fact there are special hacks to not report them as errors.
Meanwhile the lowmem mode immediately picks them as errors from day one.

The first patch fixes the original mode to report inode cache as error,
and add extra messages on how to properly fix the problem using 'btrfs
rescue clear-ino-cache'.

The second patch is a new test case for both modes of btrfs check.


check/main.c
```diff
@@ -10834,6 +10838,8 @@ static int cmd_check(const struct cmd_struct *cmd, int argc, char **argv)
 
 	ret = do_check_fs_roots(&root_cache);
 	task_stop(g_task_ctx.info);
+	if (found_free_ino_cache)
+		printf("deprecated inode cache can be removed by 'btrfs rescue clear-ino-cache'\n");
```
kdave, line 10842: please use pr_verbose(LOG_DEFAULT, ...)

check/mode-lowmem.c
```diff
@@ -5616,6 +5623,8 @@ int check_fs_roots_lowmem(void)
 
 out:
 	btrfs_release_path(&path);
+	if (found_free_ino_cache)
+		printf("deprecated inode cache can be removed by 'btrfs rescue clear-ino-cache'\n");
```
kdave, line 5627: pr_verbose

check/mode-lowmem.c
```diff
@@ -2629,6 +2630,12 @@ static int check_inode_item(struct btrfs_root *root, struct btrfs_path *path)
 		return err;
 	}
 
+	if (inode_id == BTRFS_FREE_INO_OBJECTID) {
+		error("subvolume %lld has deprecated inode cache",
```
kdave, line 2634: If this is an error then it should also exit at some point. As it is now it looks more like a warning.

kdave (approved):

check/mode-lowmem.c
adam900710: I can change that to warning, but even if we hit an error, as long as it's not critical, we should always continue other than exiting.
Or we will not fully check the whole filesystem.

PR #892, opened 2024-09-12
btrfs-progs: install btrfs-ioctl man page

maharmstone:
btrfs-ioctl.rst was laid out like it should be a man page, including having a section number, but it wasn't getting installed as such.

### DIFF
--- a/Documentation/Makefile.in
+++ b/Documentation/Makefile.in
@@ -23,6 +23,7 @@ endif
 endif
 
 mandir ?= $(prefix)/share/man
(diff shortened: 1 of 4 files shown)


kdave:
Have you checked how the manual page is rendered? IIRC the ioctl manual page was not published because it was incomplete but as it's not a stub anymore we can ship it.

maharmstone:
Yes, it looks okay to me.

kdave:
Added to devel, thanks.

PR #894, opened 2024-09-20
btrfs-progs: libbtrfsutil: use package_data member for header files

adam900710:
[BUG]
Currently with python3.12, the python bindding will always result the following warning:
```
    [PY]     libbtrfsutil
/usr/lib/python3.12/site-packages/setuptools/_distutils/extension.py:134: UserWarning: Unknown Extension options: 'headers'
  warnings.warn(msg)
```

[CAUSE]
In the setup.py which specifies the files to be included into the package, we use setuptools::Extension to specify the file lists and include paths.

But there is no handling of Extension::headers member, thus resulting the above warning.

[FIX]
Not an expert in python, but other packages like cffi are all using `package_data` to include the headers, so just follow them to use `package_data` member for headers.

Fixes: 87c3fb557ef2 ("libbtrfsutil: update python setup.py for distribution")

### DIFF
--- a/libbtrfsutil/python/.gitignore
+++ b/libbtrfsutil/python/.gitignore
@@ -5,4 +5,3 @@ __pycache__
 /build
 /constants.c
 /dist
-/pypi-README.md
--- a/libbtrfsutil/python/.gitignore
+++ b/libbtrfsutil/python/.gitignore
@@ -5,3 +5,4 @@ __pycache__
 /build
 /constants.c
 /dist
+/pypi-README.md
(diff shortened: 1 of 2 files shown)


adam900710:
Added a new patch to fix the python packaging error.

adam900710:
The existing setup.py is a disaster, as it doesn't work for venv builds.

Furthermore it doesn't follow the common python packaging directory hierarchy, it will be much better to do a rework on the build system so that at least venv build can work properly.

PR #899, opened 2024-09-27
btrfs-progs: receive: workaround full file clone quirk

adam900710:
Since kernel commit 46a6e10a1ab1 ("btrfs: send: allow cloning
non-aligned extent if it ends at i_size"), we can have clone commands to
clone a full file.

However such full file clone can have an unaligned length, and such
clone is only allowed if the destination file is no larger than the
clone range.

But we can create a case where such clone called on a destination file
which has a larger size.
This can lead to kernel reject such clone, and fail the receive.

This patchset will handle such quirk by truncate the file size to 0 if
we detect this is a full file clone.

Also add a new test case to verify the behaivor.


adam900710:
Discarded as Filipe will fix the kernel part.

PR #904, opened 2024-10-04
Fix false positives in log tree checking

maharmstone:
Filipe reported that btrfs/056 was failing due to some false positives reported by the new log-checking code in `btrfs check`.

### DIFF
--- a/check/main.c
+++ b/check/main.c
@@ -9807,9 +9807,14 @@ static int check_log_root(struct btrfs_root *root, struct cache_tree *root_cache
 			if (btrfs_file_extent_type(leaf, fi) != BTRFS_FILE_EXTENT_REG)
 				goto next;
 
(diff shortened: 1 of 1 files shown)


boryas:
btrfs/192 is broken on this as far as I can tell, and this PR helps (we get through replaying more of the FUAs) but doesn't fully fix it. Can you investigate that one as well, please? Hopefully there are still a few more kinks in the log csum checking, rather than a real btrfs bug.

fdmanana:
Btw, "sparse extents" is not valid terminology on Linux. We have sparse files, which are files with holes, and holes are regions of a file without extents.

maharmstone:
Thank you Filipe. Yes, that should probably read something like "explicit holes".

adam900710:
The 3 patches looks good to me.
But can we have some images to prevent false alerts from happening again?

PR #905, opened 2024-10-07
btrfs-progs: docs: corrected table and fix link on chunk_item.

hanyuwei70:
Corrected table format in RST.
Added minor detailed explainations.

### DIFF
--- a/Documentation/dev/On-disk-format.rst
+++ b/Documentation/dev/On-disk-format.rst
@@ -1074,27 +1074,38 @@ Off   Size Type Description
 ===== ==== ==== ==============================
 
 
-CHUNK_ITEM (e4)
+CHUNK_ITEM (0xe4)
 ^^^^^^^^^^^^^^^
 
-(100, logical address) TODO
+Key format: (100, logical address)
 
-| ``   Maps logical address to physical.``
-| ``    0  8 UINT   size of chunk (bytes)``
-| ``    8  8 OBJID  root referencing this chunk (2)``
-| ``   10  8 UINT   stripe length``
-| ``   18  8 UINT   type (same as flags for block group?)``
-| ``   20  4 UINT   optimal io alignment``
-| ``   24  4 UINT   optimal io width``
-| ``   28  4 UINT   minimal io size (sector size)``
-| ``   2c  2 UINT   number of stripes``
-| ``   2e  2 UINT   sub stripes``
-| ``   30``
-| ``   Stripes follow (for each number of stripes):``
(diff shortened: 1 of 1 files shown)


Documentation/dev/On-disk-format.rst
```diff
@@ -1074,27 +1074,38 @@ Off   Size Type Description
...
+0x0  0x8  UINT  size of chunk (bytes)
+0x8  0x8  OBJID root referencing this chunk(2)
+0x10 0x8  UINT  stripe length
+0x18 0x8  UINT  type (same as flags for block group?)
+0x20 0x4  UINT  optimal io alignment
+0x24 0x4  UINT  optimal io width
+0x28 0x4  UINT  minimal io size (sector size)
+0x2c 0x2  UINT  number of stripes
+0x2e 0x2  UINT  sub stripes (only for RAID10,fixed 2)
+0x30
```
adam900710, line 1097: I do not think we need an offset beyond the item entry.

Documentation/dev/On-disk-format.rst
```diff
@@ -1074,27 +1074,38 @@ Off   Size Type Description
...
+
+Stripes follow (for each number of stripes):
+
+==== ==== ===== ===========
+Off  Size Type  Description
+==== ==== ===== ===========
+0x0  0x8  OBJID device id
+0x8  0x8  UINT  physical offset of stripe on device
+0x10 0x10 UUID  device UUID
+0x20
```
adam900710, line 1108: The same here.

adam900710 (commented):
Overall looks good to me.

For the patch "btrfs-progs: docs: fix chunk_items section link", please also add a short commit message showing the short history of the bad link.
(e.g. introduced by which commit, and possible reason of the wrong link).

Documentation/dev/On-disk-format.rst
hanyuwei70: This is copied from original table. I kept it.
I think this is meaningful since we can use this to skip whole sector without adding last two hex number.

adam900710: Unfortunately for chunk item, we never skip the whole chunk item and go into stripe items directly.

If you really want to show the size of the item, you can add an extra size for each item.

hanyuwei70: That's a good point.
@kdave So originally what's the purpose of leaving a extra offset here?

kdave:
This has been merged already, closing. Thanks.

PR #909, opened 2024-10-15
btrfs-progs: docs: Added DEV_ITEM object id.

hanyuwei70:
Added DEV_ITEM object id to the reserved object id list. It's historical reason to let both of DEV_ITEM and ROOT_TREE have same object id. Developers should be aware of it.

### DIFF
--- a/Documentation/dev/On-disk-format.rst
+++ b/Documentation/dev/On-disk-format.rst
@@ -268,6 +268,12 @@ Otherwise, the objectid should be considered reserved for internal use.
 
    The object id that refers to the ``ROOT_TREE`` itself.
 
+-  BTRFS_DEV_ITEMS_OBJECTID = 1
+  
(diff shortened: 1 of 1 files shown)


Documentation/dev/On-disk-format.rst
```diff
@@ -1045,11 +1051,12 @@ Off   Size Type  Description
...
 
+.. _DEV_ITEM_CONTENT:
 
-DEV_ITEM (d8)
-^^^^^^^^^^^^^
+DEV_ITEM (0xd8)
+^^^^^^^^^^^^^^^
 
-(1, device id) TODO
+Key format: (1,0xd8,device id)
```
adam900710, line 1059: For the key format part, please avoid use immediate numbers, and use the `(DEV_ITEMS, DEV_ITEM, <device id>)`, just like what we did in print-tree.

adam900710: I'll change it at merge time, but if you hit similar situations, feel free to change them all.

PR #913, opened 2024-10-24
Log tree missing csum tests

maharmstone:
Adds some tests for the false positives for missing csums in the log tree, as requested by Qu in https://github.com/kdave/btrfs-progs/pull/904.


adam900710:
Thanks a lot for the new images.

Just some small nitpicks:

- You can put all the images into the same directory
  Since they are all about the same log tree check, thus a single directory is more than enough.

- Some extra dump tree output in the commit message for the involved image
  So we can have a quick view on the image.

Otherwise looks good to me.

maharmstone:
Thanks Qu, I've pushed with the changes made.

adam900710:
My bad, my previous comment is wrong, as the existing 061 is expecting the test to find errors, but the 3 new ones are expecting no false alerts.
So it's not a good idea to put the new ones into the existing directory.

Thus I have moved those to a new directory `065-valid-log-tree/` during merge.

Sorry for the extra noise.

maharmstone:
Thanks Qu. Yes, I did think that was a bit strange.

PR #919, opened 2024-11-05
btrfs-progs: refactor around btrfs_insert_file_extent()

adam900710:
Although btrfs_insert_file_extent() is a btrfs-progs specific function,
it's pretty close (although not yet similiar enough to cross-port) to
insert_reserved_file_extent() from kernel.

That kernel function uses an on-stack btrfs_file_extent_item to pass all
the needed parameters, supporting both compressed and uncompressed data
extents.

That is way more flex than btrfs_insert_file_extent() from btrfs-progs,
which can not support:

- Non-zero offset
- Ram_bytes larger than num_bytes
- Compressed extents

To prepare for the incoming support of compressed data extents
generation for mkfs --rootdir, we need a more generic way to insert file
extents.

This patch improve the situation by:

- Move btrfs_record_file_extent() to be a convert specific function
  The extra handling are all for converted btrfs, and can split extents
  where regular btrfs doesn't want.

  For mkfs/rootdir.c, the only caller out of convert, introduce a
  helper, insert_reserved_file_extent() to handle the case.

- Make btrfs_insert_file_extent() to accept an on-stack btrfs_file_extent_item
  Just like insert_reserved_file_extent() from kernel.

  Allowing us to customize ever member of the btrfs_file_extent_item.
  Now this makes btrfs_insert_file_extent() flex enough for converted
  fs, and the incoming compressed file extents.


adam900710:
Closed with all patches merged into Mark's branch.

PR #924, opened 2024-11-19
Update docstring of btrfs_util_set_default_subvolume

dcermak:
The docstring was referring to the function itself and not to the function for which it is the alias


kdave:
Merged to devel, thanks.

Issue #640, opened 2023-06-13
mkfs.btrfs refuses to make filesystem if another filesystem on the same device is mounted


DaanDeMeyer:
If I try to run mkfs.btrfs on a loop device that represents a specific offset + size of another loop device, and that other loop device has existing partitions that are mounted, mkfs.btrfs will fail with "ERROR: /dev/loopxxx is mounted". Given that it's a completely different filesystem that is actually mounted, mkfs.btrfs should not fail in this case or at least provide an option to override the check.

PR #938, opened 2025-01-14
btrfs-progs: fix overly strict is_same_loop_file()

maharmstone:
is_same_loop_file() protects us from doing mkfs.btrfs on a loopback device that's already mounted. Unfortunately it's too strict, erroring out even in some cases where mkfs would be safe. This has resulted in systemd having to do some ugly code to work around this (see https://github.com/systemd/systemd/pull/28033).

Get the offset and sizelimit of each loopback device, and allow mkfs if there's no overlap. Fixes
https://github.com/kdave/btrfs-progs/issues/640.

### DIFF
--- a/common/path-utils.c
+++ b/common/path-utils.c
@@ -124,7 +124,9 @@ static int is_loop_device(const char *device)
  * the associated file (e.g. /images/my_btrfs.img) using
  * loopdev API
  */
-static int resolve_loop_device_with_loopdev(const char* loop_dev, char* loop_file)
+static int resolve_loop_device_with_loopdev(const char* loop_dev, char* loop_file,
+					    unsigned long long *offset,
+					    unsigned long long *sizelimit)
 {
 	int fd;
 	int ret;
@@ -142,6 +144,9 @@ static int resolve_loop_device_with_loopdev(const char* loop_dev, char* loop_fil
 	memcpy(loop_file, lo64.lo_file_name, sizeof(lo64.lo_file_name));
 	loop_file[sizeof(lo64.lo_file_name)] = 0;
 
+	*offset = lo64.lo_offset;
+	*sizelimit = lo64.lo_sizelimit;
+
 out:
 	close(fd);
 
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Trying this out I get the following:

```sh
~/systemd (main)> sudo ~/.local/bin/mkosi --runtime-size=15G boot
No machine ID set, using randomized partition UUIDs.
File '/work/home/daandemeyer/systemd/build/mkosi.output/image' already is of requested size or larger, not growing. (15G >= 15G)
Applying changes to /work/home/daandemeyer/systemd/build/mkosi.output/image.
Successfully wiped file system signatures from future partition 2.
Successfully discarded data from future partition 2.
Successfully discarded gap after partition 2.
Formatting future partition 2.

btrfs-progs v6.12 
See https://btrfs.readthedocs.io for more information.

ERROR: /dev/loop5 is mounted
(mkfs) failed with exit status 1.
‣ "systemd-repart --image /work/home/daandemeyer/systemd/build/mkosi.output/image --size=16106127360 --no-pager --dry-run=no --offline=no --pretty=no /work/home/daandemeyer/systemd/build/mkosi.output/image" returned non-zero exit code 1.
```

adam900710:
Can we just simply get rid of the complex (and unnecessary to me at least) loop back device lookup?

If there are multiple loopback devices pointing to the same file, it's obviously something wrong with the end user.
And I even doubt if the existing checks is enough, e.g. using hard links to evade the path based checks.

Just treat `/dev/loop0p1` as our check boundary is good enough to me.

adam900710:
@DaanDeMeyer Mind to give this branch a try?
https://github.com/kdave/btrfs-progs/pull/940

maharmstone:
> Can we just simply get rid of the complex (and unnecessary to me at least) loop back device lookup?
> 
> If there are multiple loopback devices pointing to the same file, it's obviously something wrong with the end user. And I even doubt if the existing checks is enough, e.g. using hard links to evade the path based checks.
> 
> Just treat `/dev/loop0p1` as our check boundary is good enough to me.

This would be fine by me too.

DaanDeMeyer:
@adam900710 Will do, can you also take a look at https://lore.kernel.org/linux-fsdevel/20250115185608.GA2223535@zen.localdomain/T/#u? Your patch to remove /proc/self/fd from /proc/mounts caused a regression in systemd-repart which Boris has been investigating and that email thread has some suggestions on possible fixes.

PR #941, opened 2025-01-16
allow to print flags of subvolumes

aschnell:
Allow to print flags, so far only readonly, of subvolumes when using the 'btrfs subvolume list' command.

The use case is to avoid multiple calls to 'btrfs subvolume show' to get the information which snbk (snapper
backup tool) is doing right now.

### DIFF
--- a/Documentation/btrfs-subvolume.rst
+++ b/Documentation/btrfs-subvolume.rst
@@ -174,6 +174,8 @@ list [options] [-G [\+|-]<value>] [-C [+|-]<value>] [--sort=rootid,gen,ogen,path
                 (*parent* here means subvolume of which this subvolume is a snapshot).
         -R
                 print the UUID of the sent subvolume, where the subvolume is the result of a receive operation.
+        -F
+                print the flags of the subvolume.
 
         Type filtering:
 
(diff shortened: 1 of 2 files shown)


kdave:
The readonly is the only flag settable by the ioctl. For the `list` command update I considered also the received status, is-snapshot. If there's something more in the future sticking to the full words (`readonly`) in the output would become unfeasible. When it's used win the line output it's among the words, in the tabular output it misalignes the colums, but this is a minor problem.

```
263     1137755 256             -       @/boot/grub2/i386-pc
264     1720129 256             -       @/.snapshots
265     1720128 264             -       @/.snapshots/1/snapshot
18124   1713353 264             readonly        @/.snapshots/17866/snapshot
18125   1713355 264             readonly        @/.snapshots/17867/snapshot
```

As this is behind the option it's expected that the applications using it will understand the output, so this should be OK.

The question I still have is how to format it, this is for the original `list` command where I don't want to touch the output formatting, but this kind of minor extension is probably safe regarding compatibility. Moreover the new list command is not ready yet. I'll give it another thought and merge this PR eventually.

aschnell:
Yes, the alignment is broken. That is partly also the case for the some extra UUID options (e.g. -u) where the header is not aligned with the content. I would add a width entry in ```btrfs_list_columns``` to fix all that but I would fear breaking existing parsers.

adam900710:
If you're still interested in pushing this feature, mind to do a refresh and address the width problem mentioned by David?

aschnell:
I'm not interested anymore.

Issue #943, opened 2025-01-28
Scrub process interferes with loading and play the game and prevent shut down system


VladimirMrzv:
On my system i use HDD drive with btrfs, volume is 12 Tb. 
Every month is started scrub process but i not use pc 24x7 and when want shutdown or reboot system wait stopping process very long. For scrub I need more 20 hours and cannot ending this process.  I think it is need continius process from last block not begin from first evry time.
Also I play games and one game become very long time to load and got staters when playing. All problem from scrub, after cancel all ok. but it starts again when playing.

My system Ubuntu 24.10 and btrfs-progs v6.6.3

Forza-tng:
You can read about limiting scrub's performance imact on https://btrfs.readthedocs.io/en/latest/Scrub.html

You can cancel and resume scrubs too using `btrfs scrub cancel` and `btrfs scrub resume`.

https://btrfs.readthedocs.io/en/latest/btrfs-scrub.html

VladimirMrzv:
@Forza-tng  thanks.  Now i stop scrub before play game, but is starts again after some time. 
After reboot it starts from beginning and not resume last.

Forza-tng:
> [@Forza-tng](https://github.com/Forza-tng) thanks. Now i stop scrub before play game, but is starts again after some time. After reboot it starts from beginning and not resume last.

If it starts again it may be you have some maintenance service interfering.

VladimirMrzv:
After boot PC, scrub process begins from start. Then cancel scrub and reboot system, after boot send command to resume and it continuous scrubbing, not start again from beginning. I think it must be automatic, trying resume and only then start.

VladimirMrzv:
How I can stop scrub and run only manual?  
If I cancel process, some time ago it starts again but not resume, it starts from beginning.

VladimirMrzv:
yesterday finished scrub process, but need manual cancel before power off or reboot and resume after boot with option resume. I think need check start process and force resume not start

Heavygrass:
@VladimirMrzv 
In what version `btrfsd` appeared?

Tried
```
man btrfsd
No manual entry for btrfsd
```
I currently use 
`btrfs-progs v6.16.1`

PR #947, opened 2025-02-04
btrfs-progs: scrub: add the new -t option to set the limit at runtime

adam900710:
I know there is already `btrfs scrub limit` command to set the limit, but a lot of users are not aware of that command.

Thus adding a new option `-t <throughput_limit>` to `btrfs scrub start`.

This has some extra behavior compared to `btrfs scrub limit`:

- Only set the value for the involved scrub device(s) If it's a full fs scrub, it will be the same as `btrfs scrub limit -a -l <value>`. If it's a single device, it will bt the same as `btrfs scrub limit -d <devid> -l <value>`.

- Automatically reset the limit after scrub is finished

- It only needs one single command line to set the limit

Issue: #943
Signed-off-by: Qu Wenruo <wqu@suse.com>
---
RFC:
I'm not sure if this is really needed since we already have `btrfs scrub limit`.
Involved tools like btrfs-maintenance scripts should have such support, but to my surprise, David introduced the `btrfs scrub limit` but not adding any support to btrfs-maintenance.

### DIFF
--- a/Documentation/btrfs-scrub.rst
+++ b/Documentation/btrfs-scrub.rst
@@ -69,7 +69,7 @@ resume [-BdqrR] <path>|<device>
 
 .. _man-scrub-start:
 
-start [-BdrRf] <path>|<device>
+start [options] <path>|<device>
         Start a scrub on all devices of the mounted filesystem identified by
         *path* or on a single *device*. If a scrub is already running, the new
         one will not start. A device of an unmounted filesystem cannot be
@@ -96,6 +96,17 @@ start [-BdrRf] <path>|<device>
                 can avoid writes from scrub.
         -R
                 raw print mode, print full data instead of summary
+	--limit <limit>
+		set the scrub throughput limit for each device.
+
+		If the scrub is for the whole fs, it's the same as
+		:command:`btrfs scrub limit -a -l <value>`.
+		If the scrub is for a single device, it's the same as
+		:command:`btrfs scrub limit -d <devid> -l <value>`.
+
+		The value is bytes per second, and accepts the usual KMGT prefixes.
+		After the scrub is finished, the throughput limit will be reset to
+		the old value of each device.
         -f
                 force starting new scrub even if a scrub is already running,
                 this can useful when scrub status file is damaged and reports a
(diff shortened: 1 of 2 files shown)


kdave:
The support of limit in scrub start was "nice to have" when I added the limit command. The behaviour as you implemented it is what I'd expect.

Documentation/btrfs-scrub.rst
```diff
@@ -96,6 +96,13 @@ start [-BdrRf] <path>|<device>
                 can avoid writes from scrub.
         -R
                 raw print mode, print full data instead of summary
+	-t <throughput_limit>
```
kdave, line 99: For now please only add the long option --limit

Documentation/btrfs-scrub.rst
```diff
@@ -96,6 +96,13 @@ start [-BdrRf] <path>|<device>
...
+		set the scrub throughput limit for each device, acts the same as
+		``btrfs scrub limit -a -l <throughput_limit>``.
```
kdave, line 101: :command:`btrfs scrub limit ...`

Documentation/btrfs-scrub.rst
```diff
@@ -96,6 +96,13 @@ start [-BdrRf] <path>|<device>
...
+
+		The value is bytes per second, and accept the usual KMGT prefixes.
```
kdave, line 103: accept -> accepts

Documentation/btrfs-scrub.rst
```diff
@@ -96,6 +96,13 @@ start [-BdrRf] <path>|<device>
...
+		After the scrub is finished, the throughput limit will be reset to
+		zero (aka, no limit).
```
kdave, line 105: What if there already is some limit set, should it be reset to the original value instead?

kdave (changes requested):

Documentation/btrfs-scrub.rst
adam900710: That's a good point.

I'll save and reset to the old values instead.

kdave (approved):
Perfect, thanks.

PR #952, opened 2025-02-13
btrfs-progs: cmds/qgroup: use sysfs to detect inconsistent status early

adam900710:
Currently if "btrfs qgroup show" detects the qgroup is in an inconsistent, it will output a warning:

  WARNING: qgroup data inconsistent, rescan recommended

But the detection is based on the tree search ioctl, and qgroup tree is only updated at transaction commit time.

This means if qgroup is marked inconsistent, and the transaction is not commit, there can be a huge window as long as 30s before "btrfs qgroup show" to give a proper warning about inconsistent qgroup numbers.

To address this window, use the
'/sys/fs/btrfs/<fsid>/qgroup/inconsistent' file to do extra check.

That file is updated at real time, thus there is no delay, and can give an early warning about inconsistent qgroup numbers.

Link: https://bugzilla.suse.com/show_bug.cgi?id=1235765


kdave:
Merged to devel, thanks.

Issue #923, opened 2024-11-18
Recursive subvolume deletion only reports upper snapshot deletion.


CorrosiveTruths:
Might be a little premature, but was expecting the new recursive subvolume delete to report all the subvolumes deleted rather than just the upper level subvolume.

So with:

\# btrfs sub create butts && btrfs sub create butts/butts && btrfs sub create butts/butts/butts
Create subvolume './butts'
Create subvolume 'butts/butts'
Create subvolume 'butts/butts/butts'

Would expect something like:

\# btrfs --dry-run sub del -R butts
Delete subvolume 5896 (no-commit): '/root/btrfs-progs/butts/butts/butts'
Delete subvolume 5895 (no-commit): '/root/btrfs-progs/butts/butts'
Delete subvolume 5894 (no-commit): '/root/btrfs-progs/butts'

But you get:

\# btrfs --dry-run sub del -R butts
Delete subvolume 5894 (no-commit): '/root/btrfs-progs/butts'

realwakka:
Hi, I submitted a patch for this issue. https://lore.kernel.org/linux-btrfs/20250119105048.5452-1-realwakka@gmail.com/T/#u
I would appreciate for review. Thanks.

CorrosiveTruths:
Works as expected now, thank you.

PR #957, opened 2025-02-15
btrfs-progs: cmds: show child subvolumes during recursive delete

realwakka:
When a subvolume is deleted with the recursive option, any nested (child) subvolumes also get removed without report it. This patch modifies the delete subvol command to print a listt of child subvolumes during recursive deletion.

Issue: #923


kdave:
Thanks, merged to devel, with a test.

PR #958, opened 2025-02-16
mkfs: "defalut-ro" typo introduced in 6.13

smopucilowski:
Fixing a command line option typo in the mkfs help (`mkfs.btrfs -h`) introduced in the most recent public release: `defalut-ro` instead of `default-ro`.


kdave:
Oh the endless typos, thanks for the fix. Merged to devel.

PR #966, opened 2025-03-08
btrfs-progs: Expand abbreviations for clarity

kchr:
Replace the use of IOW as abbreviation of "In other words" with the original expanded meaning, as users who don't have English as their first language may not know what it means.


kdave:
Thanks, I'll apply the fix for the documentation. The file ctree.c is code and exposed to users, also the number of lesser known abbreviations is quite common in the code so fixing "IOW" does not improve things.

kdave:
Merged to devel.

PR #969, opened 2025-03-11
libbtrfsutil: update btrfs_util_create_snapshot doc

MikeWalrus:
The alias should be for btrfs_util_subvolume_snapshot instead of btrfs_util_snapshot_snapshot.


MikeWalrus:
Thanks.

PR #976, opened 2025-04-08
Caveats with compression and fallocate

jimis:


adam900710:
Although I have merged the two patches, they all went through some modification.
If you want to contribute to btrfs in the future, it would be better to follow the next two points:

- Write a commit message
  Even for simple doc update, please write a short description for the change.

- Use a "Signed-off-by" line
  This can be generated by "-s" option of "git commit".

kdave:
I'd like documentation updates to be easy and low barrier so I don't insist on formatting, changelog or Signed-off-by. This is also mentioned in README.md. Casual fixes or clarifications are more valuable than requiring the whole contribution process.

jimis:
Thank you! Would you consider adding a mount option for `fallocate()` to return EOPNOTSUPP and stop messing with compression?

[Our discussion](https://marc.info/?l=linux-btrfs&m=174310663519516&w=2) gives me the impression that is now a possibility:

> > Maybe the best tradeoff is to add a mount option fallocate=off.
> 
> That will be feasible.
> 
> I can try push that direction after you have updated the docs.

jimis:
I also wonder, does the case I documented here for files extended with `fallocate()`, also apply for files extended with `ftruncate()`? 

That is, such files will be excluded from future compression?

adam900710:
Only for `falloc()`, not `ftruncate()`.

The flag is added for any preallocated file extent, meanwhile truncate only creates holes.

jimis:
Thank you @adam900710. So it seems `ftruncate()` may be the most efficient way to extend a file. I'll transfer this info to the postgresql-hackers mailing list where this issue and potential workarounds are being discussed.

PR #985, opened 2025-05-14
btrfs-progs: print csum values on superblock mismatch

maharmstone:
If in btrfs_check_super() we find that the superblock has a csum mismatch, print the wanted and found values, just as we do for metadata in __csum_tree_block_size().

When hex-editing a btrfs image, it's useful to use btrfs check to calculate what the new csum should be. Unfortunately at present this only works for trees and not for the superblock, meaning you have to use the much more wordy `btrfs inspect-internal`.


kdave:
Printing the expected and found checksum makes sense but I find the reasoning a bit fishy, why do you need to hex-edit images or not want to use the existing inspect tool.

PR #987, opened 2025-05-21
btrfs-progs: new --inode-flags option

adam900710:
The new --inode-flags option allows us to specify certain btrfs specific
flags to each inode.

Currently we only support *nodatacow* and *nodatasum*.

But in the future compression flag can also be added, allowing more
accurate per-file compression.

Furthermore child inodes will inherit the flag from their parents,
meaning one only needs to specify the flag to the parent directory, then
all children files/directories will have the flag.

This new option also works well with --subvol, although one has to
note that, the inode flag inheritance does not cross subvolume boundary
(the same as the kernel).

Finally, nodatacow and nodatasum will disable compression, just like the
kernel.

I hope this will helps projects like `mkosi` and `yocto` to build their
images with more btrfs specific flags.

The only downside for now is the memory usage, both `--subvol` and
`--inode-flags` will take over 8K for each specification.
So one should not specify them too many times.

### DIFF
--- a/Documentation/btrfs-rescue.rst
+++ b/Documentation/btrfs-rescue.rst
@@ -50,6 +50,25 @@ fix-device-size <device>
 
                 WARNING: CPU: 3 PID: 439 at fs/btrfs/ctree.h:1559 btrfs_update_device+0x1c5/0x1d0 [btrfs]
 
+fix-data-checksum <device>
+	fix data checksum mismatch
+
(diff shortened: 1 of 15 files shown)


adam900710:
Thanks a lot, a new fix is sent. And if David is fine with the fix, it can be folded into the original patch (which is still only in devel branch).

kdave:
Merged to devel, thanks.

PR #988, opened 2025-05-24
btrfs-progs: convert: fix a long bug that bgt feature never works

adam900710:
There is a long bug that, "btrfs-convert -O bgt" doesn't create a fs
with bgt feature at all.

In fact "mkfs.btrfs -O bgt" is no different than "mkfs.btrfs -O ^bgt".

The root cause is explained and fixed in the second last patch.

The first 7 patches are mostly preparation and cleanup for properly
intorduce block group tree at temprory fs creation time.


kdave:
Merged to devel, thanks.

PR #990, opened 2025-05-26
btrfs-progs: minor cleanups related to extent buffer

adam900710:
The first patch is a cleanup to remove the unused @fs_info parameter
from btrfs_csum_data().

The second patch uses btrfs_csum_data() directly to calculate super
block checksum and avoid use extent buffer helpers.

The third patch removes a convert dedicated helper, and replaces it with
an existing helper.

The final patch enhance btrfs_print_leaf() to be called with temporary
extent buffers in mkfs/convert, which have no eb->leaf populated.


kdave:
Merged to devel, thanks.

PR #991, opened 2025-05-28
btrfs-progs: check that device byte values in superblock match those in chunk root

maharmstone:
The superblock of each device contains a copy of the corresponding struct btrfs_dev_item that lives in the chunk root.

Add a check that the total_bytes and bytes_used values of these two copies match.


maharmstone:
Cheers Qu, I'll have a look

maharmstone:
> But there are still some more problems need to be solved:
> 
>     * fsck/047
>       This requires proper repair support, your enhanced check didn't re-read the updated superblock after repair, thus failling that test case.
>       This need some updates on your patch.

This wasn't quite right. The problem was that the dev_rec->byte_used value wasn't getting updated in check_device_used(). I've force-pushed the patch with this change squashed in.

>     * fsck/057
>       The check is not handling seed device correctly.
>       The sprout fs can modify the old chunks on the seed device, e.g.  remove the old empty SYSTEM chunk.
>       But since the seed device is completely read-only, the device item can not be updated, thus causing check error.
>       The patch should not check a seed device against the sprouted fs.

Unfortunately this wasn't right either. btrfs-check only works on one device at a time, so if the superblock is readonly on one device the metadata on it will be too.
The problem was that this test does a mount, and it was the kernel causing the corruption. Cherry-picking 9516bae0d79045004f0b64b1f852d177cacee094 causes the problem to go away.

maharmstone:
No worries, thanks Qu

adam900710:
Considering it's still causing problems in the CI systems (the kernel is not having the backport), I'll change the check to do a warning other than an error instead.

PR #1006, opened 2025-07-25
btrfs-progs: check: add detection for missing root orphan items

adam900710:
There is an internal bug report that some half-dropped subvolume makes balance to fail.

It turns out that, the involved subvolume doesn't have an orphan item, this means the half-dropped subvolume is never going to be cleaned up, and caused problems for balance.

The problem for progs is that, neither original mode nor lowmem detects this problem.

Original mode has a bad logic which prevents us from calling the proper orphan item check.
Lowmem mode doesn't even take orphan item into consideration for subvolumes.

This series will add the detection part for btrfs-check, for both original and lowmem mode, with a hand crafted image for test.


kdave (approved):
thanks

PR #1009, opened 2025-08-02
btrfs-progs: device_get_partition_size*() enhancement and cleanup

adam900710:
[CHANGELOG]
v2:
- Include and update Zoltan's fix to device_get_partition_size_sysfs()
  The "/dev/block/<device>/size" attribute is always in sector unit
  (512Bytes), independent from the physical device's block size.

  So the fix is pretty straightforward, just do the left shift before
  returning.

  And update the commit message to properly reflect that.

- Use s64 as the return value of device_get_partition_size*()
  Kernel ensures the max sector number of a device is LLONG_MAX >> 9,
  so s64 is more than enough to cover the device size, and we are safe
  to use the minus part to pass errors.

- Extra error handling when device_get_partition_size*() failed
  Ouput an error message and exit.
  Most callers are already doing that correctly but some are not.

- Keep the sysfs method as the fallback way to grab device size
  Since the device_get_partition_size_sysfs() can be easily fixed, no
  need to use complex path search for sector size.

- Add a new cleanup to remove is_vol_small()

Zoltan recently sent a fix to solve the bug in
device_get_partition_size_sysfs(), which is causing problems for "btrfs
dev usage".

It turns out that, the bug is just the attribute
"/sys/block/<device>/size" is in sector unit (512B), not in bytes.

So fix the bug first by reporting the correct size in bytes.

Then remove several unused functions exposed during the error handling
cleanup.

Finally cleanup device_get_partition_size*() helpers to provide a proper
error handling, not just return 0 for errors, but proper minus erro
code, and extra overflow check.

### DIFF
--- a/check/main.c
+++ b/check/main.c
@@ -5447,7 +5447,13 @@ static int process_device_item(struct rb_root *dev_cache,
 				device->devid);
 			goto skip;
 		}
-		block_dev_size = device_get_partition_size_fd_stat(device->fd, &st);
+		ret = device_get_partition_size_fd_stat(device->fd, &st, &block_dev_size);
+		if (ret < 0) {
+			errno = -ret;
+			warning("failed to get device size for %s, skipping size check: %m",
+				device->name);
+			goto skip;
+		}
 		if (block_dev_size < rec->total_byte) {
 			error(
 "block device size is smaller than total_bytes in device item, has %llu expect >= %llu",
--- a/check/mode-lowmem.c
+++ b/check/mode-lowmem.c
@@ -4822,7 +4822,14 @@ next:
 			dev->devid);
 		return 0;
 	}
-	block_dev_size = device_get_partition_size_fd_stat(dev->fd, &st);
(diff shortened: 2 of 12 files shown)


kdave:
> And the original bug is that the sysfs interface only gives the device size in block unit, and block size can be hard to fetch (partition and dm have different path to grab the block size).
> 
> This makes me wonder, why we need the sysfs interface in the first place. If the unprivileged user has no access to a block device, we should not workaround it, but give a proper error/warning message.

The sysfs fallback is for the case when the device size is not readable from the block device but the size is not a secret information. Eg. it's also visible in /proc/partitions, there's no reason to hide it from the user. Forcing to use sudo for basic size information is going against usability.

Reporting the device slack needs some additional information so this should be fixed but otherwise reading the device sizes should stay as is.

adam900710:
OK, I'll drop the last patch so the sysfs fallback is still available, and allow Zotan's fix to be applied.

kdave (approved):
Ok with the s64 change discussed in the mailinglist. Thanks.

PR #1017, opened 2025-08-13
btrfs-progs: fix spurious FST entries

maharmstone:
This is something I ran into while working on https://github.com/maharmstone/btrfs-discard-check/, but Leo also ran into it last year, in some patches that seem to have been forgotten about.

When mkfs.btrfs cleans up the dummy block groups that it creates, it leaves some spurious entries behind in the free-space tree. btrfs check is supposed to diagnose this, but due to a bug it only picks these up if they're at the very end of the free-space tree.

Fix mkfs so it cleans up after itself properly, fix btrfs check so that it diagnoses the problem properly, and fix fsck/054 to get rid of the unintended spurious entries there.


adam900710:
And a fsck test case would be appreciated to make sure btrfs-check can detect such problem.

PR #1019, opened 2025-08-17
btrfs-progs: enhance --subvol/--inode-flags options

adam900710:
Currently both --subvol and --inode-flags save the full path into their
structures, and check each inode against those full path.

For long paths it can be time consuming, and this introduces extra
memory for each structure.

This series enhance the handling of those options by:

- Extract the validation part into a dedicated helper inside
  rootdir.[ch]

- Use st_dev/st_ino to replace full_path
  This reduces runtime and memory usage for the involved structures.

- Remove the memory usage warning note
  Even with the old 8K per structure memory usage, 1024 options will
  only 8M memory, that's accetable even for a lot of micro-controllers,
  not to mention modern desktop/servers.

  I'm a little paranoid at that time, with the memory usage almost
  halved, we can safely remove that warning note.

### DIFF
--- a/Documentation/mkfs.btrfs.rst
+++ b/Documentation/mkfs.btrfs.rst
@@ -243,11 +243,6 @@ OPTIONS
 	each subvolume is independent and will not inherit from the parent directory.
 	(The same as the kernel behavior.)
 
-        .. note::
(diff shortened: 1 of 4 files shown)


kdave (approved):
thanks

Issue #1015, opened 2025-08-09
Confusion over direct I/O statement in documentation


sitsofe:
Over on https://github.com/kdave/btrfs-progs/blob/3da67f1c78bcbd7556f04d89e31aa4621face50b/Documentation/ch-compression.rst#L170-L172 there is the following:

> [...] Direct IO read works on compressed files but will fall back to buffered writes and leads to no compression even if force compression is set.

I'm a little confused as to its meaning. Is it saying:
1. Direct I/O reads on compressed files bypass the page cache never fall back to buffered I/O. Separately if you try and do a Direct I/O write against a compressed file then it will fall back to a buffered write and no compression of the write will be done even force compression is set.
2. Direct I/O reads on compressed files will always work **but may fall back to buffered I/O** (it might not if the stars align!). Separately if you try and do a Direct I/O write against a compressed file then it will fall back to a buffered write and no compression of the write will be done even force compression is set.
3. Something else?

I think some of the confusion stems from the fact the sentence starts out talking about "Direct IO read" but then goes on to talk about writes.

(Mentioning @hanyuwei70, @kdave as they are listed in https://github.com/kdave/btrfs-progs/commit/f240b9f6d04100604dfb605182729f5f49657ff3)

hanyuwei70:
When I submitted this patch, I didn't read the source code but just confirmed with @adam900710 . 
I am uncertain about read, but Direct IO on write would fallback to to a  buffered write and no compression is applied even force compression is set.

PR #1020, opened 2025-08-19
btrfs-progs: docs: update the compatibility about compression

adam900710:
There is a github issue that expressed confusion about the "Compatibility" section of "ch-compression.rst".

The words in the man page is indeed confusing, and some points are no longer correct either.

Considering how complex the direct IO and compression thing is, I didn't come up with a correct answer until reading the code.

So update that section to provide a more straightforward result.

Issue: #1015


UserlandAlchemist:
It looks like there’s a minor typo:

```
Compression requires both data checksum and COW, so either *nodatasum* or
*nodatasum* mount option/inode flag will result no compression.
```

I believe the second *nodatasum* should be *nodatacow*.

kdave (approved):
Ack, thanks.

kdave:
Merged to devel, thanks.

PR #1022, opened 2025-08-22
btrfs-progs: add duplicate filename check

maharmstone:
If when doing btrfs check we encounter a DIR_ITEM slot with multiple entries due to a hash collision, loop through them to make sure that they are all distinct names.


kdave:
Merged to devel, thanks.

PR #1023, opened 2025-08-26
btrfs-progs: mkfs: add --reflink option

maharmstone:
When we are using mkfs.btrfs with --rootdir to create an image, we read each file into memory and write it back into the new image. Add an option --reflink, which uses FICLONERANGE to do this instead.

This obviously only works if the source directory and image are located on the same volume, and if the filesystem supports FICLONERANGE. We also have to handle the last sector manually if the file isn't a whole number of sectors, as the ioctl can't handle this if we're writing into the middle of a file.


kdave:
Thanks, added to devel. Please use the message helpers and the genreal of error reporting used elsewhere in the code, ie. `error` and print errno messages as `%m` instead of just the number. I've fixed it.

PR #1025, opened 2025-08-31
btrfs-progs: add dump-tree like tree-id parse support for tree-stat

SaltyKitkat:
Make the `-t` argument of tree-stat work like the one of dump-tree.


kdave:
The first patch is buggy, when the function was moved to parse utils the argument `str` was not used and left `optarg`. As optarg is global it compiles and may work by accident. Fixed in git.

PR #1032, opened 2025-09-15
btrfs-progs: docs: Fix typo

bobolopolis:
The value is called read_io_errs elsewhere.


kdave:
Merged to devel, thanks.

PR #1039, opened 2025-09-27
btrfs-progs: remove btrfs_fs_info::leaf_data_size

adam900710:
[BUG]
There is a bug report that legacy code of "btrfs rescue chunk-recover" is triggering false alerts from tree-checker, and refuse to work:

  # btrfs rescue chunk-recover /dev/nvme1n1p1
  Scanning: DONE in dev0
  corrupt leaf: root=1 block=13924671995904 slot=0, unexpected item end, have 16283 expect 0 <<< Note the "expect 0"
  leaf 13924671995904 items 11 free space 12709 generation 1589644 owner ROOT_TREE
  leaf 13924671995904 flags 0x1(WRITTEN) backref revision 1
  [...]
  Couldn't read tree root
  open with broken chunk error

[CAUSE]
The item end checks is from __btrfs_check_leaf() from tree-checker, and for the first slot of a leaf, the expected end should be BTRFS_LEAF_DATA_SIZE(), which is fetched from fs_info->leaf_data_size.

However for the fs_info opened by chunk recover, it's not going through the regular open_ctree(), but open_ctree_with_broken_chunk(), which doesn't populate that member and resulting BTRFS_LEAF_DATA_SIZE() to return 0.

[FIX]
There is no need to cache leaf_data_size, as it can be easily calulated using nodesize.

And kernel is already doing that, so follow the kernel to remove btrfs_fs_info::leaf_data_size, and use a simple inline function to do the calculation instead.

Reported-by: Mikhail Gavrilov <mikhail.v.gavrilov@gmail.com>
Link: https://lore.kernel.org/linux-btrfs/CABXGCsOug_bxVZ5CN1EM0sd9U4JAz=Jf5EB2TQe8gs9=KZvWEA@mail.gmail.com/


kdave (approved):
Thanks.

PR #1043, opened 2025-10-06
Add repair functionality for orphan device extents

adam900710:
Currently any orphan device extents will cause kernels to reject the mount.

But on the other hand, btrfs check is unable to repair such problem, so end users may fall into a situation where they are unable to mount the fs.

Add  the repair functionality to btrfs check, for both original lowmem modes, and add a new test case for it.


kdave (approved):
ack, thanks

PR #1056, opened 2025-11-02
Revert "btrfs-progs: convert: make sure the length of data chunks are also stripe aligned"

adam900710:
This reverts commit 136c58617c55358f5af7355e1847fe16edd497c0.

[BUG]
There is a bug report that kernel is rejecting a converted btrfs that has dev extents beyond device boundary.

The involved device extent is at 999627694980, length is 30924800, meanwhile the device is 999658557440.

The device is size not aligned to 64K, meanwhile the dev extent is aligned to 64K.

[CAUSE]
For converted btrfs, the source fs has all its freedom to choose its size, as long as it's aligned to the fs block size.

So when adding new converted data block groups we need to do extra alignment, but in make_convert_data_block_groups() we are rounding up the end, which can exceed the device size.

[FIX]
The original commit 136c58617c55 ("btrfs-progs: convert: make sure the length of data chunks are also stripe aligned") is only a workaround to address a kernel bug.

But since the kernel bug is long fixed, there isn't really much need to do the workaround, just revert it to avoid the dev extent to go beyond device size.

Reported-by: Andieqqq <zeige265975@gmail.com>

### DIFF
--- a/Documentation/Feature-by-version.rst
+++ b/Documentation/Feature-by-version.rst
@@ -177,7 +177,7 @@ features see :doc:`Status<Status>` page.
 6.12 (stable)
 -------------
 
--6.12 - cancellable discard/TRIM
+6.12 - cancellable discard/TRIM
         Add more points where the discard can be interrupted by signals before
         it finishes the whole operation.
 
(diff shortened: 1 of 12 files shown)


kdave (approved):
ack, thanks

kdave:
Merged to devel.

PR #1082, opened 2026-02-01
scrub: Avoid wraparound in ETA calculation

myzhang1029:
This is only a trivial patch to the scrub logging UI.

If the user deletes a significant amount of data during an ongoing scrub, `bytes_scrubbed` may exceed `bytes_total`, and it results in an unreasonable ETA estimation that might confuse users.
This patch pegs the ETA to 0 when such an underflow happens.


kdave:
Merged to devel, thanks.

PR #1087, opened 2026-02-16
btrfs-progs: add offline conversion to remap tree

maharmstone:
Add an option `--convert-to-remap-tree` to btrfstune, which converts a filesystem to use the new remap tree feature. There's no `--convert-from-remap-tree` yet, as it's considerably more complicated: it implies undoing any remappings that have already taken place.

This does the following:

* Enables the block group tree and free space tree, if these aren't already set
* Sets the remap tree incompat flag
* Removes the data reloc tree
* Creates a METADATA_REMAP block group
* Creates an empty remap tree
* Extends the block group items to their new form

### DIFF
--- a/tune/convert-remap-tree.c
+++ b/tune/convert-remap-tree.c
@@ -1,0 +1,207 @@
+/*
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public
+ * License v2 as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program; if not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+ * Boston, MA 021110-1307, USA.
+ */
+
+#include "common/messages.h"
+#include "kernel-shared/transaction.h"
+#include "kernel-shared/disk-io.h"
+#include "kernel-shared/volumes.h"
+#include "tune/tune.h"
+
+static int remove_data_reloc_tree(struct btrfs_trans_handle *trans)
+{
+	struct btrfs_fs_info *fs_info = trans->fs_info;
+	struct btrfs_root *root;
+	int ret;
+
+	struct btrfs_key key = {
+		.objectid = BTRFS_DATA_RELOC_TREE_OBJECTID,
+		.type = BTRFS_ROOT_ITEM_KEY,
+		.offset = 0,
+	};
+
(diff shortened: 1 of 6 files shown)


tune/convert-remap-tree.c
```diff
@@ -0,0 +1,197 @@
...
+
+	root = btrfs_read_fs_root(fs_info, &key);
+	if (IS_ERR(root))
+		return PTR_ERR(root);
+
+	ret = btrfs_del_root(trans, fs_info->tree_root, &root->root_key);
+	if (ret)
+		return ret;
+
+	ret = btrfs_free_tree_block(trans, btrfs_root_id(root), root->node,
```
adam900710: Why not to use `btrfs_delete_and_free_root()`? As I find this part missing the eb freeing of that function.

And maybe some checks on the level of the tree just in case.
I know you have checked that no running balance before going here, but if the tree by somehow has a lot of inode items, it can still cause problems.

tune/convert-remap-tree.c
```diff
@@ -0,0 +1,197 @@
...
+	if (ret < 0)
+		return ret;
+
+	leaf = path.nodes[0];
+
+	while (true) {
+		if (path.slots[0] >= btrfs_header_nritems(path.nodes[0])) {
+			ret = btrfs_next_leaf(fs_info->block_group_root, &path);
+			if (ret)
+				break;
```
adam900710: Missing error handling of `ret < 0` case.
In fact this break only works for `ret > 0` case.

tune/convert-remap-tree.c
```diff
@@ -0,0 +1,197 @@
...
+	ret = btrfs_search_slot(trans, fs_info->block_group_root, &key, &path,
+				-1, 1);
+	if (ret < 0)
+		return ret;
+
+	leaf = path.nodes[0];
+
+	while (true) {
+		if (path.slots[0] >= btrfs_header_nritems(path.nodes[0])) {
+			ret = btrfs_next_leaf(fs_info->block_group_root, &path);
```
adam900710: Is `btrfS_next_leaf()` safe here?

It doesn't COW the next leaf IIRC.

Thus the safest way to do this is to search the tree with COW for every block group item.
(Which can be fetched from chunk mapping or block group cache)

maharmstone:
Thank you Qu, all of that sounds correct.

I've pushed another version with your changes, would you mind having a look at that please? I also realized that it wasn't populating the remap root details in the superblock (there's no test for that yet!), so I've added another patch doing that

adam900710:
The code looks good to me now. Although please add a basic test case, with at least a newly created empty btrfs to be converted to remap tree.
Preferably with some contents populated and converted to remap tree.

But in the long run, the single large transaction solution must be changed, or it will not be able to handle huge fses.

maharmstone:
Thanks Qu, I've added a simple test

adam900710:
Thanks a lot Mark, this version looks good to me now. Merged.

PR #1089, opened 2026-02-16
btrfs-progs: device stats: drop redundant !opt_offline check

jelly:
!opt_offline is already guarded by the if statement before btrfs_open_mnt.


kdave:
Merged to devel, thanks.

PR #1091, opened 2026-02-22
Fix dropped 'const' qualifiers exposed by the latest GCC (15.2.1)

adam900710:
It looks lile the latest gcc has taken the 'const' qualifier checks one step further, that a char pointer that points into a string that has 'const' qualifier should also has 'const'.

This will expose unexpected modification like the following:

```
void my_func(const char *options)
{
        char *dot;

        dot = strchr(options, '.');
        if (!dot)
                *dot = '\0';
}
```

In above example, `dot` is either NULL or points to a location inside `options`. For the later case, since `dot` itself is not const, we can modify the content, resulting that we modify the content of `options`.
The latest GCC is able to detect such proxy modification and gives us warning on them.

And in fact, btrfs-progs has exactly such proxied modification in `bconf_save_param()`.

Other than that, most are just false alerts and we can fix them by adding a `const` quailifer.


kdave:
Merged to devel, thanks.

PR #1092, opened 2026-02-22
btrfs-progs: dump-tree: add extra chunk type checks

adam900710:
There is a report in the mailing list where a seemingly confused end user is passing a physical bytenr into "btrfs inspect dump-tree".

The best case is that bytenr has no corresponding chunk and will be rejected early.

But if there is a chunk cover that bytenr, and that chunk is a data chunk, "btrfs ins dump-tree" will mostly result a checksum mismatch, e.g.:

 $ btrfs ins dump-tree -b 13631488 ~/test.img
 btrfs-progs v6.19
 checksum verify failed on 13631488 wanted 0xcdcdcdcd found 0xf9d338a0
 ERROR: failed to read tree block 13631488

Note that 13631488 is the logical bytenr of the first data chunk.

This can be confusing for end users, as they won't know if it's really some corrupted tree blocks, or they are just passing wrong numbers.

Enhance this by introducing a new check_metadata_logical() helper, to make sure a chunk exists and the type is also correct.

Now the error would be more obvious to end users:

 $ ./btrfs ins dump-tree -b 13631488 ~/test.img
 btrfs-progs v6.19
 ERROR: logical 13631488 is not in a metadata chunk, found chunk 13631488 len 8388608 flags 0x1

Link: https://lore.kernel.org/linux-btrfs/77535c20-da3e-4dfb-b3d7-9426c25b5da3@free.fr/


kdave:
Merged to devel, thanks.

PR #1093, opened 2026-03-02
btrfs-progs: avoid using %pV

maharmstone:
The printf specifier %pV is a kernel extension that's not supported by glibc. Because of this, the various error functions in tree-checker.c all print the memory address of the va_list rather than what they're supposed to.

Change these functions so that they use the PV macros already defined in messages.h.

### DIFF
--- a/kernel-shared/tree-checker.c
+++ b/kernel-shared/tree-checker.c
@@ -84,18 +84,17 @@ static void generic_err(const struct extent_buffer *eb, int slot,
 			const char *fmt, ...)
 {
 	const struct btrfs_fs_info *fs_info = eb->fs_info;
-	struct va_format vaf;
+	DECLARE_PV(vaf);
 	va_list args;
 
 	va_start(args, fmt);
 
-	vaf.fmt = fmt;
-	vaf.va = &args;
+	PV_ASSIGN(vaf, fmt, args);
 
 	btrfs_crit(fs_info,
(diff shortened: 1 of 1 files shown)


adam900710:
Looks good to me. Merged. Thanks.

kdave:
Glibc supports that, we register the pV handler print_va_format(). The workaround is there due to musl, so what problem did you find on glibc?

maharmstone:
One of the error functions, I can't remember which, was outputting strings ending e.g. "0x7fffea62901bV". glibc 2.42

kdave:
This looks like the registration did not work, because if it goes through any of the helpers that does it then %pV works. Can you please reproduce it? 

```
checking for printf.h... yes
checking for register_printf_specifier... yes
checking for register_printf_modifier... yes
```

For musl we need to pV workaround but so far it looks like there's another problem if it does not work on glibc.

kdave:
All of the message helpers in tree-check.c call btrfs_crit which ends up in _btrfs_printk, that got the workaround in 164bc10dfc08d8754d .

maharmstone:
I think the problem is that HAVE_REGISTER_PRINTF_SPECIFIER isn't in my include/config.h, because I ran configure before 9fdb8d7469260630f674ff16b72166b924ae596a was added. %pV works on a clean checkout with glibc.

kdave:
Ok, changelog updated so it reflects the real reason for pV.

PR #1096, opened 2026-03-20
mkfs: use BTRFS_IOC_GET_CSUMS if available

maharmstone:
Use the new BTRFS_IOC_GET_CSUMS ioctl when doing mkfs.btrfs --rootdir if it's available and the csum algorithms match, so we can use the values on disk rather than spending time recomputing them.

For the case when we're also using --reflink, this means that we can now forgo reading the data entirely.

### DIFF
--- a/mkfs/rootdir.c
+++ b/mkfs/rootdir.c
@@ -39,6 +39,7 @@
 #include "kernel-lib/sizes.h"
 #include "kernel-shared/accessors.h"
 #include "kernel-shared/uapi/btrfs_tree.h"
+#include "kernel-shared/uapi/btrfs.h"
 #include "kernel-shared/extent_io.h"
 #include "kernel-shared/ctree.h"
 #include "kernel-shared/volumes.h"
@@ -160,6 +161,9 @@ static u64 default_subvol_id;
 static enum btrfs_compression_type g_compression;
 static u64 g_compression_level;
 static bool g_do_reflink;
+static int g_get_csums_supported = -1; /* -1 = unknown, 0 = no, 1 = yes */
+static dev_t g_last_csums_dev;
+static bool g_last_csums_dev_ok;
 
 static inline struct inode_entry *rootdir_path_last(struct rootdir_path *path)
 {
@@ -701,6 +705,7 @@ struct source_descriptor {
 	const char *path_name;
 	char *comp_buf;
 	char *wrkmem;
+	dev_t st_dev;
 };
 
 static int do_reflink_write(struct btrfs_fs_info *info,
@@ -915,6 +920,10 @@ static int add_file_item_extent(struct btrfs_trans_handle *trans,
 	bool datasum = true;
 	u64 flags = btrfs_stack_inode_flags(btrfs_inode);
 	off_t next;
+	u8 *fetched_csums = NULL;
+	void *fetched_csums_buf = NULL;
+	u64 fetched_num_csums = 0;
+	u16 fetched_csum_size = 0;
 
 	if (g_do_reflink || flags & BTRFS_INODE_NOCOMPRESS)
 		do_comp = false;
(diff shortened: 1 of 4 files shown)


mkfs/rootdir.c
```diff
@@ -810,6 +819,129 @@ static int add_file_item_extent(struct btrfs_trans_handle *trans,
...
+		 */
+		if (source->st_dev != g_last_csums_dev) {
+			struct btrfs_ioctl_fs_info_args fi;
+
+			g_last_csums_dev = source->st_dev;
+			memset(&fi, 0, sizeof(fi));
+			fi.flags = BTRFS_FS_INFO_FLAG_CSUM_INFO;
+
+			if (ioctl(source->fd, BTRFS_IOC_FS_INFO, &fi) == 0 &&
+			    fi.csum_type == root->fs_info->csum_type)
```
adam900710, line 857: We need to check not only the csum type but also the block size.

maharmstone: Thanks Qu, good point

mkfs/rootdir.c
```diff
@@ -973,6 +982,131 @@ static int add_file_item_extent(struct btrfs_trans_handle *trans,
...
+	 * Try to get csums from the source filesystem via BTRFS_IOC_GET_CSUMS
+	 * instead of computing them from the data.  This works when the source
+	 * is a btrfs filesystem using the same checksum algorithm, and the data
+	 * is not being compressed.
+	 *
+	 * For reflink, this is the fast path: we skip reading file data and
+	 * use the pre-computed csums directly.  For non-reflink, the data is
+	 * still read normally but we skip the checksum computation.
+	 */
+	if (datasum && !do_comp && g_get_csums_supported != 0) {
```
adam900710, line 996: This if branch is very large, it would be better to move it into a helper function.

Although the parameter/interface may need some extra consideration. We need to not only tell if we got the csum from the ioctl, but also a pointer to the allocated `*cargs`.

Maybe using something like `int try_get_csums(const struct source_descriptor *src, struct btrfs_ioctl_get_csums_args **ret_cargs)` ?

mkfs/rootdir.c
```diff
@@ -973,6 +982,131 @@ static int add_file_item_extent(struct btrfs_trans_handle *trans,
...
+		if (g_do_reflink) {
+			/*
+			 * Reflink fast path: skip reading data entirely.
+			 * Only read the trailing partial sector if needed for
+			 * do_reflink_write()'s zero-padding.
+			 */
+			to_write = ioctl_len;
+			write_buf = source->buf;
+
+			if (to_read % sectorsize) {
```
adam900710, line 1086: Why we need to bother reading the last block?

The kernel has ensured the range beyond i_size is always zeroed, and at read time we also do extra zeroing beyond i_size.
Furthermore the kernel is allowing reflink the whole inode, even if i_size is not aligned.

adam900710 (commented):
One minor issue related to whether we should extract the csum ioctl calling into a helper.

Another one related to whether we need to bother the extra handling of the last block.

Otherwise looks good to me.

mkfs/rootdir.c
Zygo: There is a gotcha case:

1. write 1M + 3K to file A
2. reflink file A to file B
3. truncate file A or file B to 1M + 2K

The truncation doesn't modify the extent or its csums, and it doesn't create a new data ref or extent in the file.  Only the file length in the inode changes.  So you can get files with some non-zero data past EOF on disk.  That non-zero data is included in the original file's csum, but won't be included in a copy of that data made by userspace.

IIRC, the kernel's read paths prevent userspace from getting the data--truncate to make the file larger triggers CoW and the copy is zero-filled; mmap can't map an incomplete block; read explicitly memsets the end of the last block before it goes to page cache.  The past-EOF data only shows up when you look at the on-disk data like this.

I ran into this while trying to use btrfs csums for dedupe.  It affects csum use cases, but it's lower level than that--if there are multiple refinks to an extent, and they are in files with different EOF tail lengths, then some of the extent's data blocks literally have different contents depending on which file you read them through, even though they're all in the same physical block on disk.  I didn't try to solve that--I just detect the condition and skip the extent.

adam900710: 3. will call `btrfs_truncate_block()` which will cause a new block.

Zygo: This is #3:

```
# head -c $((1024*1024+3072)) /dev/urandom > a
# cp --reflink a b
# truncate -s $((1024*1024+2048)) b
# sync
# btrfs-search-metadata file a
inode objectid 212489400 generation 2708934 transid 2708934 size 1051648 nbytes 1052672 block_group 0 mode 0100644 nlink 1 uid 0 gid 0 rdev 0 flags 0x0(none)
inode ref list objectid 212489400 parent_objectid 212489399 size 1
  inode ref index 2 name utf-8 'a'
extent data at 0 generation 2708934 ram_bytes 1052672 compression none type regular disk_bytenr 49942381068288 disk_num_bytes 1052672 offset 0 num_bytes 1052672
# btrfs-search-metadata file b
inode objectid 212489401 generation 2708934 transid 2708934 size 1050624 nbytes 1052672 block_group 0 mode 0100644 nlink 1 uid 0 gid 0 rdev 0 flags 0x0(none)
inode ref list objectid 212489401 parent_objectid 212489399 size 1
  inode ref index 3 name utf-8 'b'
extent data at 0 generation 2708934 ram_bytes 1052672 compression none type regular disk_bytenr 49942381068288 disk_num_bytes 1052672 offset 0 num_bytes 1052672
# cmp a b
cmp: EOF on b after byte 1050624, in line 4060
```

Same extent, same ref, same csums, different content, no COW.  If you were to copy b, the csum won't match on the receiver.

adam900710: Reading b will result 0 no matter if the on-disk extent has any other data beyond EOF.

And still you didn't explain why the extra COW is needed, you just explained it's possible to have non-zeroed data beyond EOF, that's not the affecting the data read out.

Again, reading beyond EOF will ensure it's properly zeroed, and if you want to enlarge beyond EOF, that block will still be properly zeroed.

Zygo: The problem arises when you try to use the csums from the extent directly.  The csum hasn't changed between a and b, so it includes the data from file a.  If file b is copied to the target filesystem, using the csum from file a, the last block won't match because reading file b won't give you the last 1K of data from a.

maharmstone: > Why we need to bother reading the last block?
> 
> The kernel has ensured the range beyond i_size is always zeroed, and at read time we also do extra zeroing beyond i_size.

I will have to check, but I'm pretty sure that the on-disk data between EOF and the end of the allocation isn't guaranteed to be zeroed.

> Furthermore the kernel is allowing reflink the whole inode, even if i_size is not aligned.

This is only true if you're making a copy of a file using reflinks. Here we're reflinking a file's data into a larger file (the disk image), and the size needs to be aligned for this. This was one of the surprising things that tripped me up when implementing `--reflink` in the first place... and does strongly imply that, as above, data beyond EOF isn't guaranteed to be zeroed.

adam900710: It doesn't look correct if we can read out garbage beyond EOF, there are quite some test cases verifying beyond EOF reading.

For reading a new folio, `btrfs_do_readpage()` will zero out any byte beyond EOF.

For shrinking the last block, we call `btrfs_setsize()` -> `btrfs_truncate()` -> `btrfs_truncate_block()` to handle the last block, and if needed we zero out the tailing part and dirty the block for IO later.

For expanding the last block, we call `btrfs_cont_expand()` which again calls `btrfs_truncate_block()` to do the folio zeroing.

If we can really read out garbage beyond EOF, I'd say it's a big bug and should be fixed inside the kernel.

Zygo: Again, you can't _read_ the garbage, but it is present in the _csum_, so they don't necessarily match.

That would be a vulnerability if the csum ioctl is unprivileged, but presumably it's locked down to CAP_SYS_ADMIN.

maharmstone: Userspace can't read out garbage beyond EOF, the kernel zeroes it, but there may be garbage beyond EOF on the disk. This is why reflink doesn't work for the last partial sector, and why we can't trust the csum of the last partial sector.

Zygo: Hmmm OK, so it's an unprivileged ioctl, but it filters out sensitive data (e.g. compressed csums).  ~~Hopefully it doesn't give out the csum of unaligned EOF blocks?  If not, now is the time to fix it.~~

Looks like the ioctl does filter out the EOF block:

```
+       if (!IS_ALIGNED(args.offset, fs_info->sectorsize) ||
+           !IS_ALIGNED(args.length, fs_info->sectorsize))
+               return -EINVAL;
```

So it behaves a little differently from reading the csums with TREE_SEARCH_V2, which is what I'm more familiar with.  In that case, you wouldn't get a csum from the ioctl at all, and there would be no choice but to read the tail block.

adam900710: @Zygo  Please stop making the review off-topic. If you don't know how the ioctl work, go read the kernel patch and comment on the kernel patch. All your comments are not helping but distracting the review on the progs patch.

maharmstone: With respect, Qu, Zygo's comments were interesting and helped me think about potential issues with the design of the ioctl.

adam900710: @maharmstone  Thanks, I see now. It's the limit of reflinking, that we can not reflink the full last block. Even if btrfs allows reflinking the full file.

The limit is to avoid garbage being read out, and the full file reflink ensures the dst file has the same i_size, so no way to read out the garbage.
But if reflink allows full block reflink for the last one, it can result the tailing part to be read out.

Then it's fine to handle the last block manually.

maharmstone: > @maharmstone Thanks, I see now. It's the limit of reflinking, that we can not reflink the full last block. Then it's fine to handle the last block manually.

Yes, that's right

Zygo: I admit I got some details wrong along the way, but the answer to _every_ "why do we need a weird special case at EOF on btrfs" question is always "because the truncate gotcha exists."  The details are irrelevant, because the analysis always ends at that point, no matter what intermediate steps it goes through.  Reflink can't be allowed to change EOF alignment because of the truncate gotcha.  csums at unaligned EOF can't be returned to users because of the truncate gotcha.

The gotcha can't be fixed because it's decades-old legacy behavior available to unprivileged users.  Even if the kernel did CoW in the future, files with sensitive data past EOF may exist in filesystems now, so the restrictions must remain.  Dedupe, send, clone, garbage collectors, real _or hypothetical_ csum ioctls, and reflink-cloning mkfs all have to live with it.

adam900710:
@maharmstone Considering there is only minor refactor needed, I can merge the patch and later do some refactor, and add a test case.

If you are not already working on the refactor, I'll merge the patch as is. Would that be OK for you?

maharmstone:
> @maharmstone Considering there is only minor refactor needed, I can merge the patch and later do some refactor, and add a test case.
> 
> If you are not already working on the refactor, I'll merge the patch as is. Would that be OK for you?

Absolutely. Thanks Qu

PR #1100, opened 2026-03-24
btrfs-progs: fix false alerts in lowmem mode check about verity items

adam900710:
Leo recently fixed a false alerts in btrfs-check where verity items are reported as unknown items.

Add a test case for such case, the image is generated by enabling fsverity on a small file, which will have VERITY_DESC_ITEM and VERITY_MERKLE_ITEM items to trigger the previous false alerts.

Also add a test case to catch such problems.


kdave:
Merged to devel, thanks.

PR #1103, opened 2026-03-30
btrfs-progs: mkfs/rootdir: cleanup and test case enhancement

adam900710:
[CHANGELOG]
v3:
- Remove the fiemap based prealloc detection
  The complexity doesn't really bring any other benefit and preallocated
  space itself is not that commonly used for fs image creation.

- Update the image size estimation
  To fully co-operate with the new hole detection.

v2:
- Add a new test case to verify hole detection with/without no-holes
  There is a bug in the ^no-holes handling that an incorrect file extent
  is inserted with uninitialized type.

  Exposed and fixed by Mark.

- Add a new test case to verify the file contents of hole detection

  There is a bug in the refactor of read, which overwrites the buffer
  instead of properly advance the cursor.

  Exposed by Mark with Chris Mason's review prompts.

- Keep the existing first block based bad compress ratio detection
  Previously if we had bad compression ratio after the first block, we
  do not mark the inode incompressible.

  Follow the existing behavior.

- Fix the hole size capping in fiemap mode
  Which previously is not capping the hole size, and can trigger
  UASSERT()s.

- Enhance the new fiemap test case with fssum
  To verify both the hole and file contents matches.

The first 2 patches are enhancing mkfs test cases to be more robust,
covering both no-holes and ^no-holes features, as Mark exposed a bug
affecting ^no-holes only in previous hole-detection.

The 3rd patch extracts btrfs_insert_hole_extent() to make it a little
simpler to use, without the need to populate a local on-stack file
extent item.

The 4th patch makes compressed write path easier to read, without
combining both compressed and uncompressed paths.

The final one is to use st_blocks to estimate the data size, so the
resulted fs image can be as small as possible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -553,7 +553,7 @@ test-misc: btrfs btrfs-image btrfs-corrupt-block mkfs.btrfs btrfstune fssum fsst
 	@echo "  TEST     misc-tests.sh"
 	$(Q)bash tests/misc-tests.sh
 
-test-mkfs: btrfs mkfs.btrfs
+test-mkfs: btrfs mkfs.btrfs fssum fsstress
 	@echo "  TEST     mkfs-tests.sh"
 	$(Q)bash tests/mkfs-tests.sh
 
--- a/kernel-shared/file-item.c
+++ b/kernel-shared/file-item.c
@@ -25,6 +25,23 @@
 #include "kernel-shared/file-item.h"
 #include "kernel-shared/extent_io.h"
 #include "common/internal.h"
+#include "common/messages.h"
+
+int btrfs_insert_hole_extent(struct btrfs_trans_handle *trans,
+			     struct btrfs_root *root, u64 objectid, u64 pos,
+			     u64 num_bytes)
+{
+	struct btrfs_file_extent_item stack_fi = { 0 };
+	const u32 blocksize = root->fs_info->sectorsize;
+
+	UASSERT(IS_ALIGNED(pos, blocksize));
+	UASSERT(IS_ALIGNED(num_bytes, blocksize));
+
+	btrfs_set_stack_file_extent_type(&stack_fi, BTRFS_FILE_EXTENT_REG);
+	btrfs_set_stack_file_extent_num_bytes(&stack_fi, num_bytes);
+	btrfs_set_stack_file_extent_ram_bytes(&stack_fi, num_bytes);
+	return btrfs_insert_file_extent(trans, root, objectid, pos, &stack_fi);
+}
 
 #define MAX_CSUM_ITEMS(r, size) ((((BTRFS_LEAF_DATA_SIZE(r->fs_info) - \
 			       sizeof(struct btrfs_item) * 2) / \
--- a/kernel-shared/file.c
+++ b/kernel-shared/file.c
@@ -150,7 +150,6 @@ int btrfs_punch_hole(struct btrfs_trans_handle *trans,
 		     u64 ino, u64 offset, u64 len)
 {
 	struct btrfs_path *path;
-	struct btrfs_file_extent_item stack_fi = { 0 };
 	int ret = 0;
 
 	path = btrfs_alloc_path();
@@ -165,10 +164,7 @@ int btrfs_punch_hole(struct btrfs_trans_handle *trans,
 		goto out;
 	}
 
-	btrfs_set_stack_file_extent_type(&stack_fi, BTRFS_FILE_EXTENT_REG);
-	btrfs_set_stack_file_extent_num_bytes(&stack_fi, len);
-	btrfs_set_stack_file_extent_ram_bytes(&stack_fi, len);
-	ret = btrfs_insert_file_extent(trans, root, ino, offset, &stack_fi);
+	ret = btrfs_insert_hole_extent(trans, root, ino, offset, len);
 out:
 	btrfs_free_path(path);
 	return ret;
--- a/mkfs/rootdir.c
+++ b/mkfs/rootdir.c
@@ -411,8 +411,7 @@ static int insert_reserved_file_extent(struct btrfs_trans_handle *trans,
 	 * hole.  And hole extent has no size limit, no need to loop.
 	 */
 	if (disk_bytenr == 0)
-		return btrfs_insert_file_extent(trans, root, ino,
-					       file_pos, stack_fi);
+		return btrfs_insert_hole_extent(trans, root, ino, file_pos, num_bytes);
 
 	path = btrfs_alloc_path();
 	if (!path)
@@ -819,10 +818,7 @@ static int add_file_item_extent(struct btrfs_trans_handle *trans,
 		 */
 		const u64 length = min_t(u64, next - file_pos, SZ_1G);
 
-		btrfs_set_stack_file_extent_type(&stack_fi, BTRFS_FILE_EXTENT_REG);
-		btrfs_set_stack_file_extent_num_bytes(&stack_fi, length);
-		btrfs_set_stack_file_extent_ram_bytes(&stack_fi, length);
-		ret = btrfs_insert_file_extent(trans, root, objectid, file_pos, &stack_fi);
+		ret = btrfs_insert_hole_extent(trans, root, objectid, file_pos, length);
 		if (ret < 0) {
 			error("cannot insert hole for range [%llu, %llu)",
 			      file_pos, file_pos + length);
--- a/mkfs/rootdir.c
+++ b/mkfs/rootdir.c
@@ -2048,6 +2048,24 @@ int btrfs_mkfs_fill_dir(struct btrfs_trans_handle *trans, const char *source_dir
 	return 0;
 }
 
+static void add_data_size(const char *fpath, const struct stat *st)
+{
+	int fd;
+	off_t ret;
+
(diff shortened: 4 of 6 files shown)


maharmstone:
I passed this patch series through to Claude:

● Here's my review of the four patches:

  ---
  Patch 1/4: Extract read_from_source() helper

  Generally clean refactor. A few issues:

  1. Missing buf offset in pread call. The helper reads into buf but never advances the buffer pointer:
  ret = pread(fd, buf, length - cur, filepos + cur);
  1. This should be buf + cur — without it, every iteration overwrites the beginning of the buffer and only the last chunk's data is preserved. The original code correctly
  used source->buf + bytes_read.
  2. Return type mismatch. The function returns int but returns length which is u32. This works in practice but the int return type is used for both error codes and byte
  counts — same pattern as the parent function's issue flagged in the previous review.
  3. Inline extent path change is subtly different. The old code for inline extents (around line 1334) did a single pread and checked for ret_read == -1. The new code uses the
   helper which loops until all bytes are read. This is actually a fix (partial reads were silently accepted before), but worth noting it changes behavior.

  ---
  Patch 2/4: Extract try_compressed_write()

  Good cleanup, separating the compression path. Issues:

  1. lzo_compress_extent gets first_sector instead of sectorsize. The original code at line 84:
  comp_ret = lzo_compress_extent(first_sector, source->buf, ...);
  1. The first argument to lzo_compress_extent should be blocksize (sectorsize), not first_sector (a bool). Looking at the original code before patch 1, it passed sectorsize.
  This is a regression introduced in this patch.
  2. Typo in doc comment (line 43): "submittted" → "submitted".
  3. -E2BIG handling changes semantics. The old code only set INODE_NOCOMPRESS on first_sector && -E2BIG. The new code in add_file_item_extent always sets INODE_NOCOMPRESS on
  any -E2BIG return, regardless of whether it's the first sector. This means if a later extent is incompressible, the inode gets marked NOCOMPRESS even though earlier extents
  were successfully compressed — changing the on-disk result.
  4. Redundant UASSERT(length > 0) — length is u32, always >= 0, and the subsequent min_t + <= sectorsize check handles the degenerate case.

  ---
  Patch 3/4: Use fiemap for prealloc detection

  Most significant patch, several issues:

  1. insert_hole_extent sets num_bytes/ram_bytes to 0. Lines 93-94:
  btrfs_set_stack_file_extent_num_bytes(&stack_fi, 0);
  btrfs_set_stack_file_extent_ram_bytes(&stack_fi, 0);
  1. These should be set to length, not 0. A hole extent item (for non-NO_HOLES filesystems) needs num_bytes = length to indicate the range it covers. With 0, btrfs check
  would flag this as corruption. The stack_fi is already zero-initialized, so explicitly setting to 0 is both redundant and wrong.

  1. Also, same bug as the previous commit review: type is never set to BTRFS_FILE_EXTENT_REG — it remains 0 (INLINE).
  2. UASSERT(length < INT_MAX) in insert_hole_extent but callers can pass larger values. The try_fiemap path computes length = min_t(u64, length, MAX_EXTENT_SIZE) which is
  128MiB — fine. But the SEEK_DATA hole path passes min_t(u64, next - file_pos, SZ_1G) — 1 GiB, which exceeds INT_MAX on 32-bit. Wait, INT_MAX is ~2GiB, so 1GiB is fine. But
  still, the assert is checking < INT_MAX when the return type of the caller is int — this guards the right thing but is a bit indirect.
  3. fiemap length for prealloc can exceed MAX_EXTENT_SIZE. At line 172:
  length = extent->fe_logical + extent->fe_length - filepos;
  3. This takes the full extent length from fiemap with no cap. A prealloc extent on the source could be arbitrarily large (e.g., fallocate of 1 GiB). This length is then
  passed to btrfs_reserve_extent which may fail or create an extent larger than MAX_EXTENT_SIZE (128 MiB), violating btrfs constraints.
  4. fiemap fe_logical alignment not validated. The code checks blocksize != source->blocksize to ensure the source fs has the same block size, but fe_logical and fe_length
  from fiemap may not be aligned to blocksize (fiemap returns byte-granularity offsets). Should validate alignment before using them.
  5. FIEMAP_FLAG_SYNC forces a sync on every call. This is called per-extent in a loop over the entire file. For large files with many extents, this means an fsync-equivalent
  on every iteration, which could be very slow. Consider calling it once before the loop, or dropping the flag entirely (data was already written).
  6. Memory leak on error paths. If btrfs_reserve_extent or insert_reserved_file_extent fails in the prealloc path (lines 174-188), fiemap has already been freed — that's
  fine. But if insert_hole_extent at line 143 fails, the flow is correct. OK, no leak — the free(fiemap) calls are placed correctly. Disregard.
  7. try_fiemap returns int but can return lengths up to MAX_EXTENT_SIZE (128 MiB). Same return-type-overflow concern.
  8. Stale off64_t cast at line 223. next was changed to u64 but the SEEK_DATA result is still compared with (off64_t)-1. The lseek64 return is assigned to a u64 now, so
  (off64_t)-1 becomes 0xFFFFFFFFFFFFFFFF — the comparison still works, but it's sloppy. The next initialization to file_pos at line 205 also means errno might be stale from a
  previous call when reaching the SEEK_DATA check.

  ---
  Patch 4/4: Test case

  Functional but has issues:

  1. Wrong error message at line 103: "data extent at 0 is not a hole" should be "data extent at 32K is not a hole" — copy-paste error.
  2. Test parses dump-tree output without $SUDO_HELPER. Lines 82, 87, 92, 98, 102 run "$TOP/btrfs" ins dump-tree without $SUDO_HELPER, but the device was formatted with
  $SUDO_HELPER. If the test runs as non-root, these commands will fail to open the device. Line 80 correctly uses $SUDO_HELPER but the subsequent checks don't.
  3. Fragile grep patterns. grep -A1 "EXTENT_DATA 16384" assumes the extent type line immediately follows the key line in dump-tree output. If the output format ever changes
  (e.g., additional fields), the test breaks silently.
  4. Missing check for the data content correctness. The test verifies extent types but doesn't verify that the actual data at [16K, 32K) and [80K, 84K) matches the source. A
  btrfs check is run, which helps, but doesn't validate data integrity.

adam900710:
Thanks a lot, as usual the review indeed catches several real bugs. I really need to setup the review environment.

Will get them all fixed.

adam900710:
Although there seems to be some minor false alerts:

Patch 1:
> 3. Return type mismatch. The function returns int but returns length which is u32. This works in practice but the int return type is used for both error codes and byte counts — same pattern as the parent function's issue flagged in the previous review.

This is minor, although we should enhance the return type to `ssize_t` to be extra safe, considering all modern architectures have int type no smaller than 32 bits, using int is safe enough for our usage. The extent length is never going to exceed INT_MAX.

Patch 2:

> 5. Redundant UASSERT(length > 0) — length is u32, always >= 0, and the subsequent min_t + <= sectorsize check handles the degenerate case.

It's catching length == 0 exactly.

Patch 3:
> 1 GiB, which exceeds INT_MAX on 32-bit. Wait, INT_MAX is ~2GiB, so 1GiB is fine.

The AI is confused by itself. But anyway the UASSERT() on INT_MAX is to ensure we never overflow into minus numbers.

> 5. fiemap length for prealloc can exceed MAX_EXTENT_SIZE. At line 172:

This is helping me to catch a problem in the return value of `insert_hole_extent()`, since we can pass larger value than INT_MAX, that function should following all the other functions to return the inserted length instead, so that
we will continue to insert the remaining bytes.

But the review itself is not correct, `length` is defined as u64 so it can exceed MAX_EXTENT_SIZE without any problem.

> 7. fiemap fe_logical alignment not validated.

For preallocated extents, the fe_logical must be aligned to block size inside btrfs.
Only inlined extent can have unaligned logical, but that is not handled inside `try_fiemap()`.

> 8. FIEMAP_FLAG_SYNC forces a sync on every call

I'd prefer correctness over performance, especially for this case the source fs shouldn't be under writes anyway.

maharmstone:
> Thanks a lot, as usual the review indeed catches several real bugs. I really need to setup the review environment.
> 
> Will get them all fixed.

Claude is not perfect by any means, but it's definitely worth running your patches by him beforehand. For the recent no-holes bugfix I had already run into the problem myself, but he found it too in less than a minute without any hints.

maharmstone:
Correct me if I'm wrong, but as far as I can see the only benefit this PR has over the existing lseek approach is that it detects preallocated extents. And I can't see any real use case for this: it's adding complexity for something that is unlikely ever to be useful. We now have two paths to worry about: the lseek one and the FIEMAP one, and FIEMAP itself is very complicated.

The main use case of mkfs.btrfs --rootdir is to create filesystem images, and the main use of preallocated extents is for files you're in the process of downloading... it's unclear to me why they'd ever mix. And if you did have a use case, it'd be trivial to add a call to fallocate or truncate in the bootup script of your VM.

adam900710:
>  but as far as I can see the only benefit this PR has over the existing lseek approach is that it detects preallocated extents. 

That's correct.

And I agree that the fiemap solution is indeed complex compared to lseek.

I do not have a dedicated use-case for that, the overall idea behind this is to emulate all those fs independent extent types (regular, hole, prealloc) for `--rootdir` option.

If prealloc extents is not widely used, I can drop the fiemap support and only keep the cleanup/test cases.

kdave:
Merged to devel, thanks.

PR #1105, opened 2026-04-02
btrfs-progs: fix corruption when doing mkfs.btrfs with -O ^no-holes

maharmstone:
Commit 44b83ff9 introduces a bug: when we encountered a hole in the source file, we're creating a explicit file extent.

But as we're zero-initializing the struct btrfs_file_extent_item, we're setting its type to BTRFS_FILE_EXTENT_INLINE. Fix this by forcing it to be BTRFS_FILE_EXTENT_REG.


adam900710 (approved):
Reviewed-by: Qu Wenruo <wqu@suse.com>

PR #1109, opened 2026-04-08
btrfs-progs: add missing error handling for strdup()

hobostay:
Add proper error handling for strdup() calls that were missing
NULL checks. This prevents potential NULL pointer dereferences
when memory allocation fails.

## Files Changed:
- `btrfs-map-logical.c`: output_file = strdup(optarg)
- `btrfs-sb-mod.c`: f->name = strdup(argv[i])
- `cmds/inspect.c`: sortmode = strdup(optarg)
- `common/utils.c`: *mount = strdup(ent.path)
- `common/utils.c`: param->key = strdup(key) and param->value = strdup(value)

## Context:
Similar issues were fixed in commit b6c342bf ("libbtrfs: handle memory
allocation errors when searching uuids").

This follows the same pattern of adding proper error handling for
memory allocation failures, which could otherwise lead to NULL pointer
dereferences.

### DIFF
--- a/btrfs-map-logical.c
+++ b/btrfs-map-logical.c
@@ -263,6 +263,10 @@ int main(int argc, char **argv)
 				break;
 			case 'o':
 				output_file = strdup(optarg);
+				if (!output_file) {
+					error_mem(NULL);
+					return 1;
+				}
 				break;
 			default:
 				usage(&map_logical_cmd, 1);
--- a/btrfs-sb-mod.c
+++ b/btrfs-sb-mod.c
@@ -415,6 +415,11 @@ int main(int argc, char **argv)
 		f = &spec[specidx];
 		specidx++;
 		f->name = strdup(argv[i]);
+		if (!f->name) {
+			ret = 1;
+			error_mem(NULL);
+			goto out;
+		}
 		i++;
 		if (arg_to_op_value(argv[i], &f->fop, &f->value)) {
 			ret = 1;
--- a/common/utils.c
+++ b/common/utils.c
@@ -632,6 +632,10 @@ int find_mount_fsroot(const char *subvol, const char *subvolid, char **mount)
 				goto out;
 			found = true;
 			*mount = strdup(ent.path);
+			if (!*mount) {
+				ret = -ENOMEM;
+				goto out;
+			}
 			ret = 0;
 			goto nextline;
 		}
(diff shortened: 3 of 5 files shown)


btrfs-sb-mod.c
```diff
@@ -415,6 +415,11 @@ int main(int argc, char **argv)
 		f = &spec[specidx];
 		specidx++;
 		f->name = strdup(argv[i]);
+		if (!f->name) {
+			ret = 1;
+			error_mem(NULL);
```
adam900710, line 420: CI failed to compile due to the lack of 'messages.h' include file.

common/utils.c
```diff
@@ -952,8 +956,18 @@ void bconf_add_param(const char *key, const char *value)
...
+	if (!param->key) {
+		free(param);
+		return;
+	}
+	if (value) {
 		param->value = strdup(value);
+		if (!param->value) {
+			free(param->key);
+			free(param);
+			return;
```
adam900710, line 968: Please change the return type of the function to int, and make the callers to handle the error.

This may be a big change ,thus please do it in a dedicated patch.

convert/main.c
```diff
@@ -1983,6 +1983,10 @@ int BOX_MAIN(convert)(int argc, char *argv[])
 				break;
 			case 'O': {
 				char *orig = strdup(optarg);
+				if (!orig) {
```
adam900710, line 1986: Please don't mix code with definition.

btrfs-map-logical.c
```diff
@@ -263,6 +263,10 @@ int main(int argc, char **argv)
 				break;
 			case 'o':
 				output_file = strdup(optarg);
+				if (!output_file) {
```
adam900710, line 266: We only support one output file.

Please check if `output_file` is set first, if so it means the `-o` option is provided multiple times, and this will cause memory leakage.

adam900710 (changes requested):

kdave:
The issues are easy and I'll fix them but the patch author is `Test User <test@example.com>`, and also `Signed-off-by: Claude Opus 4.6 <noreply@anthropic.com>`. I'd rather use `Assisted-by` as kernel does, but namely the patch author should be a real person, or a github name + email.

kdave:
I'll use `hobostay@users.noreply.github.com`.

PR #1111, opened 2026-04-11
btrfs-progs: reject writes if the fs has running  replace/balance

adam900710:
Btrfs-progs never has the ability to handle reloc trees, thus if a btrfs
with running balance is passed to btrfstune, btrfstune will just ignore
all the reloc trees and continue modifying the fs.

This can easily screw up the balance resume on the next RW mount and
corrupt the fs.

For dev-replace it's very similar, we can modify the metadata before the
current replace cursor, causing writes that will not be written to the
dev-replace target device and cause corruption.

Reject those fses with running replace/balance to prevent corruption for
the following tools:

- btrfstune
  All write operations will be rejected.

- btrfs check --clear-space-cache
- btrfs rescue clear-space-cache
- btrfs rescue fix-data-checksum
- btrfs rescue clear-uuid-tree
- btrfs rescue clear-ino-cache
- btrfs check --repair

Some "btrfs rescue" subcommands do not start a new transaction and
directly update the super blocks, thus they are safe from the
replace/balance.

### DIFF
--- a/Documentation/btrfs-check.rst
+++ b/Documentation/btrfs-check.rst
@@ -110,6 +110,9 @@ DANGEROUS OPTIONS
                 *--force* to give users a chance to think twice before running repair, the
                 warnings in documentation have shown to be insufficient.
 
+	.. note::
+		Please finish/cancel the running replace/balance before using this command.
+
 --init-csum-tree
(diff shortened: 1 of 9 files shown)


kdave:
Merged to devel, thanks.

PR #1112, opened 2026-04-12
btrfs-progs: fsck-tests: add the missing binary image for fsck/071

adam900710:
Surprisingly the test case fsck/071 contains no image file.

It looks like a merge error, as the patch I sent to the mailing list correctly shows the binary file:

 https://lore.kernel.org/linux-btrfs/dd3a868f33762d907a1c8bfc99c14b47c917ba7b.1770335913.git.wqu@suse.com/

Anyway, add back the missing binary file, which is created by v6.16 btrfs-progs, which has the orphan fst entries.


adam900710:
Considering this is a very small fix with a missing test image, I'll merge it after the CI finished.

PR #1113, opened 2026-04-12
btrfs-progs: check/lowmem: fix false alerts on squota features

adam900710:
The lowmem mode check doesn't recognize the new EXTENT_OWNER_REF key
type, thus it will always treat it as an unknown inline type and cause
false alerts.

Fix it by adding such types to lowmem check.

The first patch is to fix a format string error during the add of inline
backref sequence output, which breaks the indent.

The second patch is the real fix for lowmem btrfs check on squota
feature.

The last one is a new test case for fsck on squota.


kdave:
Merged to devel, thanks.

PR #1116, opened 2026-05-02
Add CRC32C SSE4.2 tests to hash-speed/-vectest

Phobos1641:
The `hash-speedtest` and `hash-vectest` tests are missing the relevant CRC32C SSE4.2 rounds.

It's useful to see what the difference in speed between CRC32C and the other algorithms are on a CPU which supports SSE4.2 but not for example PCLMUL.

### DIFF
--- a/crypto/hash-speedtest.c
+++ b/crypto/hash-speedtest.c
@@ -188,6 +188,8 @@ int main(int argc, char **argv) {
 		{ .name = "NULL-MEMCPY", .digest = hash_null_memcpy, .digest_size = 32 },
 		{ .name = "CRC32C-ref", .digest = hash_crc32c, .digest_size = 4,
 		  .cpu_flag = CPU_FLAG_NONE },
+		{ .name = "CRC32C-SSE42", .digest = hash_crc32c, .digest_size = 4,
+		  .cpu_flag = CPU_FLAG_SSE42 },
 		{ .name = "CRC32C-NI", .digest = hash_crc32c, .digest_size = 4,
 		  .cpu_flag = CPU_FLAG_PCLMUL },
 		{ .name = "XXHASH", .digest = hash_xxhash, .digest_size = 8 },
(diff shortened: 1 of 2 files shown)


kdave:
It seems that it's only Nehalem from intel cpus that is SSE 4.2 but not PCLMUL, years 2008-2010. Alternative reason to add it to the benchmark could be that Musl does not use PCLMUL due to some linkage problems so it's disabled in `crc32c_init_accel()`.

kdave:
```
# ./hash-speedtest -f crc\*
CPU flags: 0x1ff
CPU features: SSE2 SSSE3 SSE41 PCLMUL SSE42 SHA AVX AVX2
Use filter: crc*
Block size:     4096
Iterations:     100000
Implementation: builtin
Units:          CPU cycles

   NULL-MEMCPY: cycles:     29902200, cycles/i      299,    46952.426 MiB/s
    CRC32C-ref: cycles:   2404797754, cycles/i    24047,      583.834 MiB/s
  CRC32C-SSE42: cycles:    116022590, cycles/i     1160,    12101.082 MiB/s
     CRC32C-NI: cycles:   2412085610, cycles/i    24120,      582.071 MiB/s
```

Ok, so this means that Musl hasn't been using the accelerated implementation. On the same host with glibc:

```
$ ./hash-speedtest -f crc\*
CPU flags: 0x1ff
CPU features: SSE2 SSSE3 SSE41 PCLMUL SSE42 SHA AVX AVX2
Use filter: crc*
Block size:     4096
Iterations:     100000
Implementation: builtin
Units:          CPU cycles

   NULL-MEMCPY: cycles:     25439866, cycles/i      254,    55188.562 MiB/s
    CRC32C-ref: cycles:   2407791664, cycles/i    24077,      583.108 MiB/s
  CRC32C-SSE42: cycles:     46232030, cycles/i      462,    30368.561 MiB/s
     CRC32C-NI: cycles:     46040912, cycles/i      460,    30494.650 MiB/s
```

Ideally Musl should also use PCLMUL, I don't remember what was the problem but it should be fixable. Meanwhile the SSE42 is a good workaround.

kdave:
The PCLMUL workaround for Musl is not necessary anymore, so I'll enable it.

kdave:
Merged to devel, thanks.

PR #1117, opened 2026-05-05
btrfs-progs: check: fix typo in error message

bastimeyer:
Just a low-effort PR with two minor typos fixed in `print_data_backref_error()`.

I noticed this yesterday when my BTRFS rootfs got corrupted due to a bad RAM module and reading the `btrfs check` output afterwards. I'm fine btw. I had snapshot backups and could therefore re-create the FS from scratch as the repair option didn't help on this single device volume.


kdave:
Merged to devel, thanks.

PR #1119, opened 2026-05-08
include/kerncompat.h: add fallback for static_assert

codingWiz-rick:
The build fails for some uClibc toolchains with:
undefined reference to 'static_assert'

This happens because some toolchains do not define the 'static_assert'
macro in <assert.h> even when compiling with -std=gnu11. btrfs-progs
uses static_assert in the container_of macro in include/kerncompat.h.

Fix this by providing a fallback definition using _Static_assert.

### DIFF
--- a/include/kerncompat.h
+++ b/include/kerncompat.h
@@ -31,6 +31,10 @@
 #include <endian.h>
 #include <byteswap.h>
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


kdave:
Merged to devel, thanks.

PR #1121, opened 2026-05-18
btrfs-progs: mkfs/rootdir: refactor and enhancements, part 1

adam900710:
We have "--reflink" option introduced for mkfs for a while, and recently
the new get_csums ioctl is also introduced, making
add_file_item_extent() much larger and harder to maintain.

I have some concerns about the larger and larger
add_file_item_extent(), and will address some of them in the series:

- Minor mismatches on some parameter/variable names
  Use more commonly used names to replace them.

- Duplicated partial block handling
  Which not only makes add_file_item_extent() larger but also harder to
  read/maintain.

  To remove those duplicated partial block handling, the idea is to
  shrink the range to block boundary when possible.
  If we can not shrink the range anymore, disable reflink/get_csums.

  So that we do not need to handle partial blocks for reflink/get_csums.
  All the duplicated code can be removed.

- No test case for --reflink nor the latest get_csums option
  A new test case, to verify all above mentioned cases.

### DIFF
--- a/kernel-shared/extent_io.c
+++ b/kernel-shared/extent_io.c
@@ -479,7 +479,7 @@ int read_data_from_disk(struct btrfs_fs_info *info, void *buf, u64 logical,
  * Such data will be written to all mirrors and RAID56 P/Q will also be
  * properly handled.
  */
-int write_data_to_disk(struct btrfs_fs_info *info, const void *buf, u64 offset,
+int write_data_to_disk(struct btrfs_fs_info *info, const void *buf, u64 logical,
 		       u64 bytes)
 {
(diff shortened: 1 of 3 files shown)


maharmstone:
Thanks Qu. This looks good to me.
For the test, if you're interested I also sent one to xfstests along with the original version of the ioctl patch, though it needs to be fixed up for the changes we made.

adam900710:
The xfstest test case is a good idea. I'll reply to that patch.

Thanks a lot!

adam900710:
@maharmstone  Now the series is merged. Feel free to add your reviewed-by tags in devel branch.

maharmstone:
I don't have write access unfortunately!

PR #1123, opened 2026-05-22
btrfs-progs: fix a heap-overflow bug in raid56 reconstruction

adam900710:
During my local tests on a kernel patchset of falling back to
RAID1/RAID1C3 for single-data-RAID56, it turns out that "btrfs check
--check-data-csum" can report a lot of false csum mismatch.

It turns out that there is heap-overflow bug where we can get garbage
into the IO buffer for RAID56 reconstruction read.

Fix it and add a regression test case.


kdave:
Queued for devel, thanks.

PR #1125, opened 2026-05-29
btrfs-progs: fix the overly strict chunk type check on dump-tree

adam900710:
Commit 408e45feb325 ("btrfs-progs: dump-tree: add extra chunk type
checks") introduced a chunk type check before dumping the tree.

However it's too strict as I'm a total idiot, and forgot to we still
need to dump chunk tree blocks.

Fix it and add a regression test.


kdave:
Queued for devel, thanks.

PR #1126, opened 2026-05-29
mkfs.btrfs: support reproducible image creation

loemraw:
Make `mkfs.btrfs` produce byte-for-byte identical images from identical
inputs (same binary, args, and source tree), for supply-chain verification,
build caching, and meaningful image diffs.

Block placement was already deterministic; this series plugs the remaining
non-determinism sources.

## Interface

No new CLI flags — two environment variables (same convention as the
reproducible-builds ecosystem and xfsprogs):

- `SOURCE_DATE_EPOCH=<unix_ts>` — pins all timestamps.
- `DETERMINISTIC_SEED=1` — opt into deterministic UUID derivation (requires a
  fixed `-U`; errors out if missing instead of silently emitting random UUIDs).

The `-U` UUID is the seed; all other internal UUIDs derive from it.

```
chown -R 0:0 ./rootfs              # caller normalizes ownership
truncate -s 1G img.btrfs           # fresh, zeroed target
SOURCE_DATE_EPOCH=1700000000 DETERMINISTIC_SEED=1 \
    mkfs.btrfs -f -U <uuid> -r ./rootfs img.btrfs
```

## What's fixed (one per commit)

- `time(NULL)` root-item timestamps → `SOURCE_DATE_EPOCH`.
- Random UUIDs (chunk-tree, FS_TREE, per-device, per-subvol) → derived from the
  fs UUID as `SHA256(fs_uuid || role || key)`, kept unique for
  multi-device/multi-subvol correctness.
- Host leaks: `nr_global_roots` (CPU count) and `BIG_METADATA` (page size) made
  host-independent.
- `-r` source file `atime/ctime/mtime` → `SOURCE_DATE_EPOCH` (ctime can't be
  normalized externally).
- `-r` directory order: `nftw()` (filesystem-dependent) replaced with a
  name-sorted walker, so the same tree on ext4/xfs/tmpfs yields the same image.

Ownership is left to the caller (`chown`) since the value is image-semantic.

## Testing

Includes `tests/reproducibility-check.sh` (root-required). Verified
byte-identical across ext4-vs-xfs source, multi-device, all csum types,
nodesize/sectorsize/compression/features, special files/xattrs/ACLs, allocator
perturbation, ASLR, and TZ/locale/umask/cwd.

## Caveats

- Target must be fresh/zeroed (mkfs doesn't scrub free space, like mkfs.ext4).
- `--compress` output also depends on the compression library version.

### DIFF
--- a/Documentation/mkfs.btrfs.rst
+++ b/Documentation/mkfs.btrfs.rst
@@ -708,6 +708,48 @@ by large metadata blocks and space reservation strategy that allocates more
 than can fit into the filesystem.
 
 
+ENVIRONMENT
+-----------
+
+:command:`mkfs.btrfs` honors two environment variables that make image
+creation reproducible, so that the same inputs produce a byte-for-byte
+identical image.
+
+SOURCE_DATE_EPOCH
+        A decimal count of seconds since the Unix epoch
+        (`reproducible-builds.org <https://reproducible-builds.org>`_),
+        used in place of the current time for the timestamps written into
+        the image: the root-item and inode times of the created trees.
+        With *--rootdir*, each copied file's *ctime* and *atime* are set to
+        this value. Its *mtime* is taken from the source but clamped to it,
+        so nothing in the image is newer than the source date.
+
+        An empty value is treated as unset. A value that is not a
+        non-negative integer fitting in the platform time type is a hard
+        error.
+
+DETERMINISTIC_SEED
+        When set to *1*, the internal UUIDs that are otherwise random (the
+        chunk-tree UUID, each device UUID and each subvolume UUID) are
+        instead derived deterministically from the filesystem UUID. A fixed
+        filesystem UUID must be supplied with *-U*; without it
+        :command:`mkfs.btrfs` exits with an error rather than emit a random
+        image. Any value other than *1* leaves UUID generation random.
+
+To create a reproducible image, pin the filesystem UUID with *-U*, set
+``SOURCE_DATE_EPOCH`` and ``DETERMINISTIC_SEED``, normalize source ownership,
+and start with a fresh target file. :command:`mkfs.btrfs` copies *uid* and
+*gid* from the source tree and, like other mkfs tools, does not zero space it
+does not write:
+
+.. code-block:: bash
+
+        $ chown -R 0:0 ./rootdir
+        $ truncate -s 1G image.btrfs
+        $ SOURCE_DATE_EPOCH=1700000000 DETERMINISTIC_SEED=1 \
+                mkfs.btrfs -U <uuid> --rootdir ./rootdir image.btrfs
+
+
 AVAILABILITY
 ------------
 
--- a/common/path-utils.c
+++ b/common/path-utils.c
@@ -20,7 +20,9 @@
 #include <linux/kdev_t.h>
 #include <linux/loop.h>
 #include <linux/limits.h>
+#include <dirent.h>
 #include <fcntl.h>
+#include <ftw.h>
 #include <unistd.h>
 #include <stdlib.h>
 #include <stdio.h>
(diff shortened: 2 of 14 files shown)


common/path-utils.c
```diff
@@ -495,3 +497,187 @@ char *path_dirname(char *path)
...
+		if (!de)
+			break;
+		if (strcmp(de->d_name, ".") == 0 ||
+		    strcmp(de->d_name, "..") == 0)
+			continue;
+		if (count == cap) {
+			struct path_walk_entry *new_entries;
+
+			cap = cap ? cap * 2 : 16;
+			new_entries = realloc(entries,
```
boryas: this is reallocarray

common/path-utils.c
```diff
@@ -495,3 +497,187 @@ char *path_dirname(char *path)
...
+					      cap * sizeof(*entries));
+			if (!new_entries) {
+				ret = -ENOMEM;
+				goto cleanup;
+			}
+			entries = new_entries;
+		}
+		entries[count].name = strdup(de->d_name);
```
boryas: I think this is fine cause the dirent.name is null-terminated, but I think it has a bound so you could consider using strndup. no big deal, to me.

common/path-utils.c
```diff
@@ -495,3 +497,187 @@ char *path_dirname(char *path)
...
+		size_t name_len = strlen(entries[i].name);
+
+		/* Need room for "/" + name + terminator after path_len. */
+		if (path_len + 1 + name_len + 1 > path_buf_size) {
+			ret = -ENAMETOOLONG;
+			goto cleanup;
+		}
+		path[path_len] = '/';
+		memcpy(path + path_len + 1, entries[i].name, name_len + 1);
+
```
boryas: common/path-utils.c has a path_cat_out that I think does what you want.

common/reproducible.c
```diff
@@ -42,3 +45,38 @@ time_t reproducible_now(void)
...
+		error("SHA-256 failed deriving deterministic UUID");
+		exit(1);
+	}
+
+	memcpy(out, digest, BTRFS_UUID_SIZE);
+	/* RFC 4122 v4: set version (high nibble of byte 6) and variant
+	 * (high two bits of byte 8) so the result identifies as v4. */
+	out[6] = (out[6] & 0x0F) | 0x40;
+	out[8] = (out[8] & 0x3F) | 0x80;
+}
```
boryas, line 76: Two thoughts:
1. It was not immediately obvious to me why the interface is so complex. presumably if we are fixing a "determinstic seed" then the random uuid should just be deterministic. Some discussion explaining uuid_generate, or the xfs precedent, or generally just some high level explanation of the spec/expectation/behavior would be helpful. My GUESS is that the issue is that uuid_generate is a library function that is not configurable to have a fixed seed, so we have to sha shuffle the single well known uuid passed in. But it would be nice to make that (or the actual right answer if that is BS) clearer.
2. Some quick googling (https://man7.org/linux/man-pages/man3/uuid_generate.3.html) turned up `uuid_generate_sha1` which purports to produce a valid v5 uuid from an input uuid using sha. What is the reason to seemingly roll our own rather than use that?

boryas (commented):
Mostly looks good to me. Some little nit picks and a few questions here and there. Would be good to just do a general once over on the string processing stuff to make sure that it follows the same most common conventions in the rest of btrfs-progs.

Thanks!

mkfs/main.c
```diff
@@ -2028,8 +2035,14 @@ int BOX_MAIN(mkfs)(int argc, char **argv)
...
-	/* This is also fixed in kernel, but the flag has no real meaning anymore. */
-	if (nodesize > sysconf(_SC_PAGE_SIZE))
+	/*
+	 * This is also fixed in kernel, but the flag has no real meaning
+	 * anymore. Set it based on the historical 4K page-size assumption
+	 * rather than the build host's actual page size, so that mkfs run
+	 * on a 16K-page host (e.g. aarch64) with the default 16K nodesize
+	 * produces the same superblock bytes as mkfs on a 4K-page host.
+	 */
+	if (nodesize > SZ_4K)
```
maharmstone: I'd just take the if out and do this unconditionally. The versions of the kernel that don't support BIG_METADATA are absolutely ancient.

maharmstone:
Leo, you might want to grab a Docker image of a distro that uses musl, and check that a version of mkfs.btrfs compiled there produces a file with the same hash.

Likewise it might be an idea to do the same in a s390x VM, to double-check that you're not making any endian assumptions anywhere.

common/reproducible.c
loemraw: Your guess is right, uuid_generate can't be seeded, so we derive each UUID from the seeded fs UUID using the role and key to make each one distinct. You're also right that we should just switch to uuid_generate_sha1, no real good reason to roll our own.

mkfs/main.c
loemraw: Yeah, agreed.

maharmstone:
> Leo, you might want to grab a Docker image of a distro that uses musl, and check that a version of mkfs.btrfs compiled there produces a file with the same hash.
> 
> Likewise it might be an idea to do the same in a s390x VM, to double-check that you're not making any endian assumptions anywhere.

Don't necessarily think you need to do these things, btw, this is just me thinking out loud about what it would take to make this absolutely bulletproof.

Two more such ideas (again, you don't need to do these):
- Should we be maybe normalizing the filenames to UTF-8 and a specific normalized form? Otherwise you could have filenames in a rootdir that look the same to the user but create different images
- What happens if read() gets interrupted by a signal? If we get a different-sized read as a result that might change where the extent boundaries are

IIUC the images here are only guaranteed to be the same for the same version of btrfs-progs (and the compression libraries presumably). It's a much harder problem, but if we could guarantee that everything was well-defined and deterministic we could make it so the test was just checking the hash of the image against a precomputed value.

common/path-utils.c
```diff
@@ -495,3 +497,142 @@ char *path_dirname(char *path)
...
+	if (ret)
+		goto cleanup;
+
+	/*
+	 * readdir() returns NULL at both end-of-stream and error; clear
+	 * errno before each call to tell them apart after the loop.
+	 */
+	while (1) {
+		errno = 0;
+		de = readdir(dir);
```
adam900710, line 573: I'm not a big fan of this, we need to read out the whole directory, before doing anything, and this is the new default?
I think it's practical for building small fs images, but only for that.
What if the target directory have millions entries? I know it will definitely expose some other problems, but I prefer one less problem, not more.

Please, only do this heavy sorting when needed. Add an option to trigger the reproducible routines.

loemraw: Good call, made it opt-in in v2. Normal mkfs -r just uses nftw() like before, the sorted walk only kicks in when you're actually building a reproducible image (both SOURCE_DATE_EPOCH and DETERMINISTIC_SEED set).

adam900710: Sounds great. Thanks for the update.

adam900710:
This version looks good to me.

However I'd prefer someone from Meta to give it an extra review. As I do not have a use-case here.

Documentation/mkfs.btrfs.rst
```diff
@@ -708,6 +708,51 @@ by large metadata blocks and space reservation strategy that allocates more
...
+        When set to *1*, the internal UUIDs that are otherwise random (the
+        chunk-tree UUID, each device UUID and each subvolume UUID) are
+        instead derived deterministically from the filesystem UUID. A fixed
+        filesystem UUID must be supplied with *-U*; without it
+        :command:`mkfs.btrfs` exits with an error rather than emit a random
+        image. Any value other than *1* leaves UUID generation random.
+
+To create a reproducible image, pin the filesystem UUID and both
+variables, and normalize ownership of the source tree yourself (the one
+image attribute left to the caller, a single :command:`chown`):
```
boryas: this bit (and the truncate bit below) feel a bit vague and out of place.

It feels like you want to give a "recipe" for avoiding some gotchas, so I think doing that more totally is best. Alternatively, we can rely on the test as the documentation for the canonical way to do it. I am fine with either but this section of two comments and a code snippet incorporating only one of them doesn't make a ton of sense to me.

common/path-utils.c
```diff
@@ -508,3 +510,142 @@ int path_readlink(char *dest, const char *src)
 	dest[ret] = 0;
 	return ret;
 }
+
+/* Sorted depth-first directory walker. See path-utils.h for the contract. */
```
boryas: not that helpful of a comment IMO.

common/path-utils.c
```diff
@@ -508,3 +510,142 @@ int path_readlink(char *dest, const char *src)
...
+	struct stat st;
+	struct FTW ftwbuf;
+	struct path_walk_entry *entries = NULL;
+	size_t count = 0, cap = 0;
+	DIR *dir = NULL;
+	struct dirent *de;
+	const char *slash;
+	int ret;
+
+	/* nftw() reports base as the offset of the last path component. */
```
boryas: without a comment anywhere nearby explaining that we are emulating nftw this is sort of weird to me.

common/path-utils.c
```diff
@@ -508,3 +510,142 @@ int path_readlink(char *dest, const char *src)
...
+	dir = NULL;
+
+	if (count)
+		qsort(entries, count, sizeof(*entries), path_walk_entry_cmp);
+
+	for (size_t i = 0; i < count; i++) {
+		/*
+		 * Heap-allocate to keep a PATH_MAX buffer off the per-level
+		 * stack frame.
+		 */
```
boryas: relatively vacuous comment as well

common/reproducible.c
```diff
@@ -57,6 +57,17 @@ bool reproducible_is_deterministic(void)
...
 }
 
+bool reproducible_is_enabled(void)
+{
+	/*
+	 * A byte-reproducible image needs every source of nondeterminism
+	 * pinned: timestamps (SOURCE_DATE_EPOCH) and UUIDs (DETERMINISTIC_SEED,
+	 * which also requires -U). With only one set the image still varies,
+	 * so the extra cost of the sorted -r walk only pays off when both are.
+	 */
```
boryas: this feels redundant and tautological

mkfs/rootdir.c
```diff
@@ -2185,7 +2226,15 @@ int btrfs_mkfs_fill_dir(struct btrfs_trans_handle *trans, const char *source_dir
 	g_do_reflink = do_reflink;
 	INIT_LIST_HEAD(&current_path.inode_list);
 
-	ret = nftw(source_dir, ftw_add_inode, 32, FTW_PHYS);
+	/*
+	 * The sorted walker reads a whole directory into memory before
+	 * descending, which only pays off when a reproducible image is
```
boryas: this is kind of weird wording too, it sort of bakes in the review history. The point is that the reproducible mode is costly, not whether or not sorting "pays off"

I would say it more like: reproducible builds requires sorted traversal, which incurs additional costs. Don't incur the allocation and cpu cost of sorting for regular non-reproducible mkfs.

boryas (approved):
LGTM. Left a number of inline doc/comment nits but it looks improved since my last review and I was already pretty happy with it. I like the split sorted/nftw design as well, for what it's worth.

loemraw:
Hey, v3 is ready now. The main change from v2 is how rootdir atime is handled. v2 preserved source atimes older than SOURCE_DATE_EPOCH and only clamped newer ones, like mtime. That isn't reproducible when reading the source updates atime between builds, so v3 always pins atime to SOURCE_DATE_EPOCH. ctime remains pinned there too, while mtime still preserves older source values and clamps newer ones. I added a btrfs-backed regression for this, tightened the reproducible-image recipe in the docs, and trimmed the comments and commit messages down.

PR #1128, opened 2026-06-03
Various fixes reported by code scanning tools

kdave:
- **btrfs-progs: use path API for path building**
- **btrfs-progs: harden path building and limit length to PATH_MAX**
- **btrfs-progs: receive: handle more cases of incomplete stream**
- **btrfs-progs: use safe array realloc helper for dynamic array**
- **btrfs-progs: reset capacity if array allocation fails**
- **btrfs-progs: verify new array size before appending new item**
- **btrfs-progs: accept NULL for the array API**
- **btrfs-progs: fix memory leak on error path in check_space_cache()**
- **btrfs-progs: fix check_arg_type() to recognize UUIDs**
- **btrfs-progs: fix minor memory leak of device path in btrfs_add_to_fsid()**
- **btrfs-progs: format-output: add lower bounds for indentation**
- **btrfs-progs: use unsigned size type in check_mounted_where()**
- **btrfs-progs: validate new subvolume length**
- **btrfs-progs: send-stream: check size before reading in read_cmd()**
- **btrfs-progs: send-stream: handle EINTR during buffer reads**
- **btrfs-progs: use safer allocation in get_fs_info()**

### DIFF
--- a/common/array.c
+++ b/common/array.c
@@ -86,7 +86,7 @@ int array_append(struct array *arr, void *element)
 	if (arr->length == arr->capacity) {
 		void **tmp;
 
-		tmp = realloc(arr->data, (arr->capacity + alloc_increment) * sizeof(void *));
+		tmp = reallocarray(arr->data, arr->capacity + alloc_increment, sizeof(void *));
 		if (!tmp)
 			return -1;
 		arr->data = tmp;
--- a/common/array.c
+++ b/common/array.c
@@ -40,8 +40,10 @@ int array_init(struct array *arr, unsigned int capacity)
 		arr->capacity = alloc_increment;
 
 	tmp = calloc(arr->capacity, sizeof(void *));
-	if (!tmp)
+	if (!tmp) {
+		arr->capacity = 0;
 		return -1;
+	}
 	arr->data = tmp;
 	return 0;
 }
--- a/common/array.c
+++ b/common/array.c
@@ -88,6 +88,9 @@ int array_append(struct array *arr, void *element)
 	if (arr->length == arr->capacity) {
 		void **tmp;
 
+		if (arr->capacity + alloc_increment < arr->capacity)
+			return -1;
+
 		tmp = reallocarray(arr->data, arr->capacity + alloc_increment, sizeof(void *));
 		if (!tmp)
 			return -1;
--- a/common/array.c
+++ b/common/array.c
@@ -51,6 +51,8 @@ int array_init(struct array *arr, unsigned int capacity)
 /* Free internal data array. */
 void array_free(struct array *arr)
 {
+	if (!arr)
+		return;
 	free(arr->data);
 	arr->length = 0;
 	arr->capacity = 0;
@@ -62,6 +64,8 @@ void array_free_elements(struct array *arr)
 {
 	unsigned int i;
 
+	if (!arr)
+		return;
 	for (i = 0; i < arr->length; i++) {
 		free(arr->data[i]);
 		arr->data[i] = NULL;
@@ -72,6 +76,8 @@ void array_free_elements(struct array *arr)
 /* Reset all elements to NULL up to capacity. */
 void array_clear(struct array *arr)
 {
+	if (!arr)
+		return;
 	for (unsigned int i = 0; i < arr->capacity; i++)
 		arr->data[i] = NULL;
 }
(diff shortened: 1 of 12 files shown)


common/path-utils.c
```diff
@@ -430,7 +430,7 @@ int path_cat_out(char *out, const char *p1, const char *p2)
 		p1_len--;
 	if (p2_len && p2[p2_len - 1] == '/')
```
adam900710, line 431: Not related to this patch, but the ending '/' detection is not correct.

E.g. p1 is "whatever/////", then all the tailing "/" should be truncated, not only the last one.

Secondly, the path length check should be based on the final p1/p2 length.

Thirdly if the p1_len + p2_len + 2 matches PATH_MAX, it's still valid.

common/path-utils.c
```diff
@@ -450,7 +450,7 @@ int path_cat3_out(char *out, const char *p1, const char *p2, const char *p3)
 		p2_len--;
 	if (p3_len && p3[p3_len - 1] == '/')
 		p3_len--;
```
adam900710, line 452: The same applies to path_cat3_out().

common/array.c
adam900710, line 32: The only caller of the function is inside tests.

All real users uses on-stack zeroed memory.

Not sure if we even need this function at all.

There are even some that are defined but never utilized functions:

- array_clear()
- array_use_capacity()

common/array.c
```diff
@@ -51,6 +51,8 @@ int array_init(struct array *arr, unsigned int capacity)
 /* Free internal data array. */
 void array_free(struct array *arr)
 {
+	if (!arr)
```
adam900710, line 54: Since every real users of array is using on-stack memory, I'd prefer it to be an UASSERT() instead of handling NULL pointers.

common/root-tree-utils.c
```diff
@@ -136,7 +136,8 @@ int btrfs_link_subvolume(struct btrfs_trans_handle *trans,
 	u32 imode;
 	int ret;
 
-	UASSERT(namelen && namelen <= BTRFS_NAME_LEN);
+	if (namelen <= 0 || namelen > BTRFS_NAME_LEN)
+		return -EINVAL;
```
adam900710, line 140: Please add an error message at least.

The callers are convert (using a fixed subvolume name) or mkfs/rootdir.

The later can got the subvolume name from a directory from the host fs.
Thus if there is really something wrong, we want to notify the end user on why.

common/root-tree-utils.c
adam900710, line 128: Also I think we should use unsigned namelen.

common/path-utils.c
kdave: The path manipulation does not do any sanitization or simplification (like removing ".."), the helpers are there namely to stop strcpy+strcat or sprintf.

common/array.c
kdave: It's there for API completeness and it's not used everywhere it could. This is supposed to replace manual arrays passed as pointer and length.

common/root-tree-utils.c
kdave: Agreed, I have WIP with replacing more ints with size_t where length is used and passed as parameter.

common/root-tree-utils.c
kdave: Ok, mesage added.

common/path-utils.c
adam900710: Yes, I know. But the second and third points are still something worthy fixing.

common/array.c
adam900710: I'm fine for now since you're just fixing the obvious ones reported by whatever tools.

But if you plan to follow up, I'd prefer to have unnecessary interfaces removed.

common/root-tree-utils.c
adam900710: OK, we can leave those for the next cleanup.

adam900710:
Some mentioned points can be fixed in a dedicated series.

Otherwise the series looks good to me.

common/root-tree-utils.c
kdave: I forgot to push the branch with updated message before you merged it, I'll add it to devel.

common/array.c
kdave: I'd like to keep the APIs separate, UASSERT pulls kerncompat, stack tracing and such. For the lowest level APIs I'd rather use plain assert() or at least split common/messages.h to more independent parts.

PR #1129, opened 2026-06-09
Fixes, API updates and cleanups

kdave:
Assorted fixes, some helper and API updates to avoid hard coded behaviour.

- **btrfs-progs: use size_t for length in various places**
- **btrfs-progs: delete unused test_issubvolname()**
- **btrfs-progs: use sizeof() in snprintf() limits if possible**
- **btrfs-progs: handle potential memory allocation failure in print_one_fs()**
- **btrfs-progs: quota status: handle mode detection on kernels 6.7**
- **btrfs-progs: add sysfs file read with whitespace-clean return buffer**
- **btrfs-progs: quota status: read cleaned string from sysfs**
- **btrfs-progs: read clean exclusive op sysfs file**
- **btrfs-progs: use calloc() in get_fs_info_offline()**
- **btrfs-progs: use reallocarray() for resizing arrays**
- **btrfs-progs: use safe reallocarray() pattern in cmd_inspect_list_chunks()**
- **btrfs-progs: unify use of realloc() and reallocarray() patterns**
- **btrfs-progs: build: restore configure check for reallocarray()**
- **libbtrfsutil: remove reallocarray() stub**
- **btrfs-progs: build: add more configure check macros**
- **btrfs-progs: ci: mention Alpine in musl builds**
- **btrfs-progs: ci: fixup Leap 16 build name reference**
- **btrfs-progs: add path helper for readlink**
- **use path_readlink in add_symbolic_link**


common/sysfs-utils.c
```diff
@@ -202,3 +203,35 @@ int sysfs_write_fsid_file_u64(int fd, const char *name, u64 value)
 	close(fd);
 	return ret;
 }
+
+int sysfs_read_fsid_file_clean_str(int fd, const char *name, char *buf, size_t size)
```
adam900710, line 207: Please add a comment explaining the function and its parameter.

The parameter is pretty complex, `fd` is only utilized to get the fsid, meanwhile `name` is the filename inside the fsid directory.

And we also have a local `fd` for the sysfs file, which can be very confusing.

convert/source-reiserfs.c
```diff
@@ -165,8 +165,8 @@ static int insert_cached_objectid(reiserfs_filsys_t fs, u64 objectid)
 	struct reiserfs_convert_info *info = fs->fs_vp;
 
 	if (info->used_slots + 1 >= info->alloced_slots) {
-		u64 *objectids = realloc(info->objectids,
-				    (info->alloced_slots + 1000) * sizeof(u64));
+		u64 *objectids = reallocarray(info->objectids,
+					      info->alloced_slots + 1000, sizeof(u64));
```
adam900710, line 169: White space errors?

common/path-utils.c
```diff
@@ -482,3 +482,29 @@ char *path_dirname(char *path)
...
+ * Return number of bytes written to @dest, or negative error.
+ */
+int path_readlink(char *dest, const char *src)
+{
+	int ret;
+
+	/* Readlink does not add the final null byte and can fill the whole buffer. */
+	ret = readlink(src, dest, PATH_MAX);
+	if (ret < 0)
+		return ret;
```
adam900710: It should return `-errno` other than `ret`, as `readlink()` failure will always return -1.

mkfs/rootdir.c
```diff
@@ -358,18 +358,13 @@ static int add_symbolic_link(struct btrfs_trans_handle *trans,
 	int ret;
 	char buf[PATH_MAX];
 
-	ret = readlink(path_name, buf, sizeof(buf));
-	if (ret <= 0) {
+	ret = path_readlink(buf, path_name);
+	if (ret < 0) {
+		errno = ret;
```
adam900710: return value is minus, making errno not working for the printf.

common/sysfs-utils.c
```diff
@@ -202,3 +203,35 @@ int sysfs_write_fsid_file_u64(int fd, const char *name, u64 value)
...
+
+	ret = sysfs_read_file(fd, buf, size);
+	if (ret < 0)
+		goto out;
+
+	end = strlen(buf);
+	/* Remove trailing space (newlines). */
+	while (end >= 1) {
+		if (isspace(buf[end - 1])) {
+			end--;
```
adam900710, line 224: if `end` reached 0, we break the loop without doing any string trimming.

mkfs/rootdir.c
```diff
@@ -358,18 +358,13 @@ static int add_symbolic_link(struct btrfs_trans_handle *trans,
 	int ret;
 	char buf[PATH_MAX];
 
-	ret = readlink(path_name, buf, sizeof(buf));
-	if (ret <= 0) {
+	ret = path_readlink(buf, path_name);
```
adam900710, line 361: Commit title lacks "btrfs-progs: " prefix.

adam900710:
I also find some minor problems in previous merge. And I'll fix them in place. So please refresh the base when updating the series.

common/sysfs-utils.c
kdave: Yeah, API documentation is in the todo. I've copied the fd reuse pattern from other functions but yes it can be separate variable so it's not confusing.

convert/source-reiserfs.c
kdave: Not intentional, vim does that and trying to fix it with unidented line still inserts the spaces.

kdave: Or maybe it was wrong previously, mixing spaces with tabs, I redid it manually with just tabs and alignment with spaces, it looks like on the diff.

mkfs/rootdir.c
kdave: Right, the top commit in 'tig' shows the branch name so I overlooked lack of the prefix.

kdave:
Fix @subvol leak during error path

In the first patch "btrfs-progs: use path API for path building", the bug was already there. I think it's better not to mix such changes into a multi-file series that's targeted to fix a different problem.

adam900710:
Nope, it's exactly the patch "btrfs-progs: use path API for path building" introducing the new leakage.

Before that we just call `snprintf()` followed by `free(subvol)`, there was no leak in the first place.
It's the new error handling of `path_cat_out()` caused a new leak on `subvol`, which is new.

common/sysfs-utils.c
kdave: Which means the string is just whitespace. So with "if (end == 0) buf[0] = 0;" after the first loop it works. We don't have empty files in sysfs but it still needs to be handled.

kdave:
Ah right, so obvious now.

kdave:
Update pushed:
- return -errno
- errno = -ret
- trimming whitespace-only string handled
- subject of last patch

Issue #1124, opened 2026-05-22
btrfs filesystem mkswapfile creates unusable swapfile on 16K page systems, also -v/-q flags don't work


eConnah:
On a 16K page kernel (ARM), `mkswapfile` completes without error but the resulting swapfile can't be activated:

```
$ sudo swapon swapfile
swapon: swapfile: swap format pagesize does not match. (Use --fixpgsz to reinitialize it.)
swapon: swapfile: swapon failed: Invalid argument
```

It seems like `mkswap` is being called without `-p` so it defaults to 4096 regardless of the system page size, but there is no flag to override it. You can call `mkswap -p 16384` manually but having the flag integrated into this wrapper would be quite nice.

Also noticed that `-v` and `-q` are listed in the help output but don't actually work:

```
$ btrfs filesystem mkswapfile -v swapfile
btrfs filesystem mkswapfile: invalid option 'v'
$ btrfs filesystem mkswapfile -q swapfile
btrfs filesystem mkswapfile: invalid option 'q'
```

kdave:
`mkswap` is not called internally (ie. no option to pass), the swapfile signature is simple and is created directly. Send as PR #1130 . The created swapfile is recognized by 'file' with correct page size, I haven't tested it on real machine with `swapon`.

eConnah:
Ah sorry I didn't check properly, thank you so much for the quick fix I just ran your pr on my laptop and it worked perfectly.

Autodetect:
```fish
connor@lenix ~/D/r/btrfs-progs (pr-1130)> sudo ./btrfs filesystem mkswapfile -s 512M test_swap
create swapfile test_swap size 512.00MiB (536870912) page size 16384
connor@lenix ~/D/r/btrfs-progs (pr-1130)> sudo swapon test_swap
connor@lenix ~/D/r/btrfs-progs (pr-1130)> swapon --show
NAME                                               TYPE SIZE  USED PRIO
/swap/swapfile                                     file  12G 36.8M   -1
/home/connor/Documents/repos/btrfs-progs/test_swap file 512M    0B   -1
```

Manual
```fish
connor@lenix ~/D/r/btrfs-progs (pr-1130)> sudo ./btrfs filesystem mkswapfile -s 512M -p 16K test_swap
create swapfile test_swap size 512.00MiB (536870912) page size 16384
connor@lenix ~/D/r/btrfs-progs (pr-1130)> sudo swapon test_swap
[...]

PR #1130, opened 2026-06-10
Add page size support to mkswapfile

kdave:
- **btrfs-progs: mkswapfile: new parameter to specify page size**
- **btrfs-progs: tests: add test case for mkswapfile**

Issue #1124


adam900710:
As I explained in the original issue, the usage of swapfile on btrfs has a lot of hidden burdens, and that burden is way larger than one would expect.

Thus I prefer not to make it any easier to use swapfile on btrfs.

PR #1132, opened 2026-06-24
Support the experimantal encryption feature

dvacek:
This patch set complements the kernel changes and allows for wider testing.
These changes are needed for fstests


adam900710:
Merged the latest version from the ML.

Thanks.

PR #1133, opened 2026-06-26
btrfs-progs: skip unused tree loading for btrfs restore

SaltyKitkat:
Add granular OPEN_CTREE_SKIP_*_TREE flags allowing callers to skip loading specific trees they don't need. btrfs restore only requires chunk_root, tree_root, and fs_root -- all other trees (extent, csum, dev, free space) are loaded but never used.

Set skip flags in restore's open_fs() to avoid accessing potentially corrupted tree blocks on damaged filesystems.

Additionally fix two gaps in OPEN_CTREE_PARTIAL handling:
- load_global_roots_objectid: btrfs_search_slot failure on tree_root was always fatal, even with PARTIAL set. Handle with PARTIAL by jumping to synthetic root creation.
- btrfs_setup_all_roots: dev_root loading failure was always fatal, even with PARTIAL set. Follow the uuid_root pattern -- kfree and set to NULL on failure when PARTIAL is set.

### DIFF
--- a/kernel-shared/disk-io.c
+++ b/kernel-shared/disk-io.c
@@ -796,7 +796,7 @@ struct btrfs_root *btrfs_read_fs_root(struct btrfs_fs_info *fs_info,
 	if (location->objectid == BTRFS_CHUNK_TREE_OBJECTID)
 		return fs_info->chunk_root;
 	if (location->objectid == BTRFS_DEV_TREE_OBJECTID)
-		return fs_info->dev_root;
+		return fs_info->dev_root ? fs_info->dev_root : ERR_PTR(-ENOENT);
 	if (location->objectid == BTRFS_CSUM_TREE_OBJECTID)
 		return btrfs_global_root(fs_info, location);
 	if (location->objectid == BTRFS_UUID_TREE_OBJECTID)
@@ -999,7 +999,8 @@ static int read_root_or_create_block(struct btrfs_fs_info *fs_info,
 			error("could not setup %s tree", str);
 			return -EIO;
 		}
-		warning("could not setup %s tree, skipping it", str);
+		warning("could not setup %s tree, skipping (partial open)",
+			str);
 		/*
 		 * Need a blank node here just so we don't screw up in the
 		 * million of places that assume a root has a valid ->node
@@ -1044,8 +1045,15 @@ static int load_global_roots_objectid(struct btrfs_fs_info *fs_info,
 
 	ret = btrfs_search_slot(NULL, tree_root, &key, path, 0, 0);
 	if (ret < 0) {
-		error("could not find %s tree", str);
-		return ret;
+		if (!(flags & OPEN_CTREE_PARTIAL)) {
+			error("could not find %s tree", str);
+			return ret;
+		}
+		warning("could not find %s tree, skipping (partial open)",
+			str);
+		btrfs_release_path(path);
+		ret = 0;
+		goto create_synthetic;
 	}
 	ret = 0;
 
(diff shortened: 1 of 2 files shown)


adam900710:
I think the new `OPEN_CTREE_SKIP_*_TREE` is a little over-killed.

There is only very limited users of the old `PARTIAL` flag, mostly:
- dump-tree
- zero-log
- clear-uuid-tree
- fix-dev-size
- restore

For those operations, they do not require that small granular either.

I'd prefer to merge those new tree skips into the existing PARTIAL flag instead.

Mind to share on which cases do the new `OPEN_CTREE_SKIP_*_TREE` bring extra benefit?

SaltyKitkat:
> I think the new `OPEN_CTREE_SKIP_*_TREE` is a little over-killed.
> 
> There is only very limited users of the old `PARTIAL` flag, mostly:
> - dump-tree
> - zero-log
> - clear-uuid-tree
> - fix-dev-size
> - restore
> 
> For those operations, they do not require that small granular either.
> 
> I'd prefer to merge those new tree skips into the existing PARTIAL flag instead.

Make sense.

> Mind to share on which cases do the new `OPEN_CTREE_SKIP_*_TREE` bring extra benefit?

I've met a btrfs whose dev root (and of course a lot of other things) is broken and seems level mismatch and transaction mismatch. IMO this kind of error is common when flush or something is unsound. In this case a rescue mount seems work properly, but btrfs utils cannot work, including restore and dump-tree. It would be very helpful if they can work and print some info. I'm sure that restore don't use those trees, but not sure about how dump-tree works currently.

I think when working on a known broken fs, we should load as few things as possible to make tools more possible to work and get some useful info. the --chunk-root really helps a lot in this case because the chunk root is also broken, but because of that option is missing in some utils, I have to fix the raw bytes of chunk root before going further and try rescue mount to save data.

SaltyKitkat:
I wonder if there's any tool to edit an offline btrfs image manually to fix corrupt fs or simulate a specific corruption. That would help a lot in my case.

adam900710:
OK, totally agree on the idea to load as few trees as possible.
In that case I will be very happy to accept the patch that avoid dev and global tree loading for restore.

For the tool to corrupt/edit a btrfs, `btrfs-corrupt-block` is your friend.
It's not provided in most distros, but included and compiled by default inside btrfs-progs.

SaltyKitkat:
> I think the new `OPEN_CTREE_SKIP_*_TREE` is a little over-killed.
> 
> There is only very limited users of the old `PARTIAL` flag, mostly:
> 
>     * dump-tree
> 
>     * zero-log
> 
>     * clear-uuid-tree
> 
>     * fix-dev-size
> 
>     * restore
> 
> 
> For those operations, they do not require that small granular either.

Yes, but I found that `btrfs check --repair` also set the partial flag [here](https://github.com/kdave/btrfs-progs/blob/devel/check/main.c#L10814). So I wonder if ignore those trees is the correct behavior for `btrfs check --repair`.

adam900710:
> Yes, but I found that `btrfs check --repair` also set the partial flag [here](https://github.com/kdave/btrfs-progs/blob/devel/check/main.c#L10814). So I wonder if ignore those trees is the correct behavior for `btrfs check --repair`.

I'm less concerned about the check situation. The point is, if a tree is corrupted, with PARTIAL we can ignore it.
But if that tree root is not corrupted, even with PARTIAL, we can still load the tree.

PARTIAL doesn't mean only loading the minimal amount of trees, but allows us to ignore *bad* optional trees.

Issue #1135, opened 2026-06-27
Regression in "mkfs.btrfs --compress zlib:9 --shrink", creates broken files since v7.0 (Input/output error on certain files)


mid-kid:
To reproduce, start with an empty directory. This was tested on gentoo.

Prepare a 20MiB file:
```
# mkdir -p root
# dd if=/dev/urandom of=root/file.bin count=20MiB
```

Install an older version of `mkfs.btrfs`:
```
# emerge -1g '<sys-fs/btrfs-progs-7'
[...snip...]
# emerge --info sys-fs/btrfs-progs
[...snip...]
sys-fs/btrfs-progs-6.19.1::gentoo was built with the following:
USE="convert man zstd -experimental -reiserfs -static -static-libs -udev -verify-sig" ABI_X86="(64)"
CFLAGS="-O2 -pipe -march=x86-64 -mtune=generic -g -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing"
CXXFLAGS="-O2 -pipe -march=x86-64 -mtune=generic -g -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing"
LDFLAGS="-Wl,-O1 -Wl,--as-needed -Wl,-z,pack-relative-relocs -g -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing"
# mkfs.btrfs --version
mkfs.btrfs, part of btrfs-progs v6.19.1
-EXPERIMENTAL -INJECT -STATIC +LZO +ZSTD -UDEV +FSVERITY +ZONED CRYPTO=builtin
```

Create a disk image using this version:
```
# mkfs.btrfs -r root --compress zlib:9 --shrink good.img
```

Install a newer version of `mkfs.btrfs`:
```
# emerge -1g sys-fs/btrfs-progs
[...snip...]
# emerge --info sys-fs/btrfs-progs
[...snip...]
sys-fs/btrfs-progs-7.0::gentoo was built with the following:
USE="convert man zstd -experimental -reiserfs -static -static-libs -udev -verify-sig" ABI_X86="(64)"
CFLAGS="-O2 -pipe -march=x86-64 -mtune=generic -g -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing"
CXXFLAGS="-O2 -pipe -march=x86-64 -mtune=generic -g -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing"
[...]

PR #1136, opened 2026-06-27
btrfs-progs: fix a bug in compression ratio failed path

adam900710:
Issue: #1135 

Commit c9c8d64acd17 ("btrfs-progs: mkfs/rootdir: factor out compressed
write path from add_file_item_extent()") left two lines which still set
compression value for the uncompressed file extent.

This will lead to a corrupted file extent item, preventing users from
reading the data.

The first patch fixes the bug, and the second one is a regression test
case for it.


adam900710:
Merged even though there is no other reviewer.

The bug fix is small, and with a proper regression test.

PR #1138, opened 2026-06-29
btrfs-progs: check: add option to skip qgroup accounting verification

pdzhu:
Qgroup accounting verification can be expensive because it walks extents and backrefs to recompute referenced/exclusive numbers.

Add --skip-qgroup-accounting so btrfs check can still load qgroup status/info/relation items from the quota tree without verifying the accounting numbers.

Reject the option with --repair and --qgroup-report, as both depend on full accounting verification results.

### DIFF
--- a/Documentation/btrfs-check.rst
+++ b/Documentation/btrfs-check.rst
@@ -58,6 +58,16 @@ SAFE OR ADVISORY OPTIONS
         This expects that the filesystem is otherwise OK, and is basically an offline
         *scrub* that does not repair data from spare copies.
 
+--skip-qgroup-accounting
+        skip verification of qgroup referenced/exclusive accounting.
+
+        Qgroup status, info and relation items are still loaded from the quota
+        tree. This option cannot be used with *--repair* or *--qgroup-report*.
+
+        This can significantly reduce runtime on filesystems with many snapshots
+        or reflinked files, as full qgroup accounting verification can require
+        extensive backref walks.
+
 --chunk-root <bytenr>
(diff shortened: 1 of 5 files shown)


adam900710:
Are you hitting a case where qgroup is causing problem (too long runtime)?

If so, have you any benchmark of "btrfs check" with and without qgroup verification on that fs?

pdzhu:
Yes, the original case was a 4T filesystem where full qgroup accounting verification took hours, while skipping it reduced the check to minutes.

I did not keep the exact numbers from that run, so I'll rerun the benchmark on the same filesystem and post the results here.

Documentation/btrfs-check.rst
```diff
@@ -58,6 +58,11 @@ SAFE OR ADVISORY OPTIONS
         This expects that the filesystem is otherwise OK, and is basically an offline
         *scrub* that does not repair data from spare copies.
 
+--skip-qgroup-accounting
+        Skip verification of qgroup referenced/exclusive accounting.
```
adam900710: It would be better to add a short description that this option can greatly reduce runtime if the fs has a lot of snapshots/reflinked files.

adam900710 (approved):
Overall looks good to me, only a minor doc update needed.

adam900710:
OK, that info is more than enough. No need for a comprehensive runtime comparison. The existing commit message is enough.

Documentation/btrfs-check.rst
pdzhu: Updated the documentation as suggested. Thanks for the review.

PR #1142, opened 2026-07-06
btrfs-progs: mkfs: dedupe hardlinks when estimating rootdir size

### LINKED ISSUE #1141: mkfs.btrfs overcounts hardlinks
`btrfs_mkfs_size_dir()` overestimates the required device size when the source directory (`--rootdir`) contains hardlinked files, because it counts each hardlink's `st->st_blocks` individually.

Test example:

```sh
# Create a 10 MB source directory with 100 hardlinks to the same file
mkdir /tmp/rootdir
dd if=/dev/urandom of=/tmp/rootdir/bigfile bs=1M count=10
for i in $(seq 1 100); do
    ln /tmp/rootdir/bigfile /tmp/rootdir/hardlink-$i
done
du -sh /tmp/rootdir

# Now try mkfs.btrfs on a 1G file
truncate -s 1G /tmp/test.img
mkfs.btrfs -r /tmp/rootdir /tmp/test.img
ls -hl /tmp/test.img  # grew to ~1.28G

# With --shrink, it shrinks back down:
truncate -s 1G /tmp/test2.img
mkfs.btrfs --shrink -r /tmp/rootdir /tmp/test2.img
ls -hl /tmp/test2.img  # shrunk to ~184 MB
```

Downstream issues: systemd/mkosi#4391, systemd/systemd#42891

daradib:
Previously, ftw_add_entry_size() counted every hard link to a file as separate data, greatly overestimating the required device size for directories with many hard links to the same inode.

Track visited inodes (by st_dev/st_ino) in an rbtree during the size-estimation walk, and skip accounting for inodes already seen. Only inodes with st_nlink > 1 are checked, since single-link files cannot be duplicates.

Add a test that creates 100 hard links to a 10 MB file and verifies that mkfs.btrfs --rootdir does not grossly overestimate the required image size.

Fixes #1141.

### DIFF
--- a/mkfs/rootdir.c
+++ b/mkfs/rootdir.c
@@ -206,8 +206,6 @@ static int hardlink_compare_nodes(const struct rb_node *node1,
 
 	entry1 = rb_entry(node1, struct hardlink_entry, node);
 	entry2 = rb_entry(node2, struct hardlink_entry, node);
-	UASSERT(entry1->root);
-	UASSERT(entry2->root);
 
 	if (entry1->st_dev < entry2->st_dev)
 		return -1;
@@ -2213,6 +2211,8 @@ fallback:
 static int ftw_add_entry_size(const char *fpath, const struct stat *st,
 			      int type, struct FTW *ftwbuf)
 {
+	int ret;
+
 	/*
 	 * Failed to read the directory, mostly due to EPERM.  Abort ASAP, so
 	 * we don't need to populate the fs.
@@ -2220,9 +2220,26 @@ static int ftw_add_entry_size(const char *fpath, const struct stat *st,
 	if (type == FTW_DNR || type == FTW_NS)
 		return -EPERM;
 
-	if (S_ISREG(st->st_mode))
-		add_data_size(fpath, st);
 	ftw_meta_nr_inode++;
+
+	if (S_ISREG(st->st_mode) && st->st_size > 0) {
+		/*
+		 * Skip hard links whose backing inode has already been
+		 * accounted for to avoid overestimating the data size.
+		 */
+		if (st->st_nlink > 1) {
+			ret = add_hard_link(NULL, 0, st);
+			if (ret == -EEXIST)
+				return 0;
+			if (ret < 0) {
+				errno = -ret;
+				error("failed to add hard link record for '%s': %m",
+				       fpath);
+				return ret;
+			}
+		}
+		add_data_size(fpath, st);
+	}
 
 	return 0;
 }
@@ -2252,6 +2269,7 @@ u64 btrfs_mkfs_size_dir(const char *dir_name, u32 sectorsize, u64 min_dev_size,
 	 * follow them here.
 	 */
 	ret = nftw(dir_name, ftw_add_entry_size, 10, FTW_PHYS);
+	rb_free_nodes(&hardlink_root, free_one_hardlink);
 	if (ret < 0) {
 		error("ftw subdir walk of %s failed: %m", dir_name);
 		exit(1);
--- a/tests/mkfs-tests/044-rootdir-hardlink-size/test.sh
+++ b/tests/mkfs-tests/044-rootdir-hardlink-size/test.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+#
+# Test that "mkfs.btrfs --rootdir" correctly estimates the device size when the
+# source directory contains many hardlinks to the same file.  Without inode
+# deduplication in the size-estimation pass, each hardlink was counted
+# individually, greatly overestimating the required size and causing the output
+# file to grow unnecessarily.
+
+source "$TEST_TOP/common" || exit
+
+check_prereq mkfs.btrfs
+check_prereq btrfs
+
+# Create the rootdir with one 10 MB file and 100 hardlinks to it.
+rootdir="$(_mktemp_dir mkfs-rootdir-hardlink-size)"
+run_check fallocate -l 10M "${rootdir}/bigfile"
+for i in $(seq 1 100); do
+	run_check ln "${rootdir}/bigfile" "${rootdir}/hardlink-$i"
+done
+
+# Create a test image sized to 1 GiB.
+prepare_test_dev 1G
+
+# First test with --shrink: verify the result is well below 1G.
+run_check "$TOP/mkfs.btrfs" --shrink --rootdir "$rootdir" "$TEST_DEV"
+run_check "$TOP/btrfs" check "$TEST_DEV"
+
+img_size="$(run_check_stdout stat --format=%s "$TEST_DEV")"
+sb_total="$(run_check_stdout "$TOP/btrfs" inspect-internal dump-super "$TEST_DEV" |
+	awk '/^[[:space:]]*total_bytes[[:space:]]/{print $2; exit}')"
+
+if [ "$img_size" -ne "$sb_total" ]; then
+	_fail "shrink: image size $img_size != total_bytes $sb_total"
+fi
+if [ "$img_size" -ge $((1 * 1024 * 1024 * 1024)) ]; then
+	_fail "shrink: image size $img_size >= 1 GiB, shrink did not work"
+fi
+
+# Reuse the device at the shrunk size for the non-shrink test.
+# If the size estimation correctly deduplicates hardlinks, the data should fit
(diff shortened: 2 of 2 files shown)


mkfs/rootdir.c
```diff
@@ -134,6 +134,69 @@ struct hardlink_entry {
...
+		return 1;
+	if (e1->st_ino < e2->st_ino)
+		return -1;
+	if (e1->st_ino > e2->st_ino)
+		return 1;
+	return 0;
+}
+
+/*
+ * Returns 1 if this is the first time we see the inode (and records it),
```
adam900710: Considering `rb_insert()` can return `-EEXIST`, and this function can also return `-ENOMEM`, I'd prefer not to use >0 to indicate a proper new entry.

Just return `-EEXIST` to indicate an existing one, and return 0 for a successful insert, which is more common inside btrfs-progs.

mkfs/rootdir.c
```diff
@@ -2216,13 +2279,33 @@ static void add_data_size(const char *fpath, const struct stat *st)
...
+	 * Skip hard links whose backing inode has already been accounted for.
+	 * nftw visits every directory entry, so without deduplication a file
+	 * with N hard links would have its data (and inode metadata) counted N
+	 * times, greatly overestimating the required device size.
+	 *
+	 * st_nlink == 1 means the inode has a single name, so it cannot be a
+	 * hard link we have already seen; skip the rbtree lookup in that case.
+	 * Directories cannot be hardlinked, so we can skip them as well.
+	 */
+	if (!S_ISDIR(st->st_mode) && st->st_nlink > 1) {
```
adam900710: I'd prefer `S_ISREG()` over `!S_ISDIR()`. The later can match more things like char/bdev files, which should not take any on-disk space anyway.

Also, if the file is empty, we should skip the `size_inode_insert_new()` too.

tests/mkfs-tests/044-rootdir-hardlink-size/test.sh
```diff
@@ -0,0 +1,82 @@
...
+# file to grow unnecessarily.
+
+source "$TEST_TOP/common" || exit
+
+check_prereq mkfs.btrfs
+check_prereq btrfs
+
+tmp=$(_mktemp_dir mkfs-rootdir-hardlink-size)
+rootdir="$tmp/rootdir"
+fs_img="$tmp/test.img"
```
adam900710: It may not be  a good idea to use `$tmp` for large files, as on a lot of distros, `$tmp` is created inside `/tmp` which is tmpfs, backed by memory. There is no guarantee the test machine can hold such large file.

You can use `prepare_test_dev()` with a size, which will utilize the file inside `$TEST_TOP`, and should have more space available for most cases.

mkfs/rootdir.c
```diff
@@ -2216,13 +2279,33 @@ static void add_data_size(const char *fpath, const struct stat *st)
...
+		ret = size_inode_insert_new(st);
+		if (ret < 0)
+			return ret;
+		if (ret == 0)
+			return 0;
```
adam900710: Even if it's a hard link, it still contributes to the metadata usage. So we should only skip `add_data_size()`, but keep the later `ftw_meta_nr_inode++;` handling.

mkfs/rootdir.c
```diff
@@ -134,6 +134,69 @@ struct hardlink_entry {
 
 static struct rb_root hardlink_root = RB_ROOT;
 
+/*
+ * Track inodes already accounted for during the size estimation pass
+ * (btrfs_mkfs_size_dir), so that hard links to the same inode are only
+ * counted once. Keyed by the host filesystem's (st_dev, st_ino).
+ */
+struct size_inode_entry {
```
adam900710: Forgot to mention, there is already a structure recording every file whose st_nlink > 1:

`hardlink_entry`

You can resuse that structure and related functions during the `nftw()` inside `btrfs_mkfs_size_dir()`.

Just don't forget to clean it up before the real `nftw()` call with `ftw_add_inode()`.

daradib: Okay, I trimmed it down to reuse that structure with a placeholder for btrfs_root (`size_dir_root`) and btrfs_ino (`0`).

mkfs/rootdir.c
daradib: I modified the existing condition to be `if (S_ISREG(st->st_mode) && st->st_size > 0)` so empty files (hard link or not) are excluded from data size calculation.

tests/mkfs-tests/044-rootdir-hardlink-size/test.sh
daradib: I changed it to use the prepare_test_dev helper. The image is sparsely allocated and should be relatively small when the hard links are correctly deduplicated, but this is more consistent. Also reordered the checks so shrink runs first, then reuses the shrunk device size to verify the non-shrink path does not grow the image.

adam900710:
Please fold [btrfs-progs: mkfs: reuse hardlink_root for size estimation](https://github.com/kdave/btrfs-progs/commit/e77df7418a85ea8b82c5dd67c5aaedce2395a632) into [btrfs-progs: mkfs: dedupe hardlinks when estimating rootdir size](https://github.com/kdave/btrfs-progs/commit/cec4df2120949c741e9eb3224dc0c9a89caad1d3).

Not sure if you're LLM or not, but it makes no sense to introduce something, then remove it in the next commit.
I see no reason why a sane person will do things like this.

The same applies to the test case.

In short, please make sure every commit make sense on its own.

daradib:
I rebased the commits so there is one commit for the change to mkfs and one for the test case.

Some projects prefer to avoid rewriting commit history that have been reviewed. Apparently LLMs are getting us to impersonate them though instead of the other way around.

adam900710:
Preserving git history only makes sense for master/main branch, development branch is always fine to be force pushed.

mkfs/rootdir.c
```diff
@@ -134,6 +134,9 @@ struct hardlink_entry {
 
 static struct rb_root hardlink_root = RB_ROOT;
 
+/* Root for hard link entries stored during directory size estimation. */
+static struct btrfs_root size_dir_root;
```
adam900710: Just use NULL, there is no need for `root` usage during size estimation, nor for a dedicated pointer.

mkfs/rootdir.c
```diff
@@ -2213,17 +2216,36 @@ static void add_data_size(const char *fpath, const struct stat *st)
...
-		add_data_size(fpath, st);
 	ftw_meta_nr_inode++;
 
+	if (S_ISREG(st->st_mode) && st->st_size > 0) {
+		/*
+		 * Skip hard links whose backing inode has already been
+		 * accounted for to avoid overestimating the data size.
+		 */
+		if (st->st_nlink > 1) {
+			ret = add_hard_link(&size_dir_root, 0, st);
```
adam900710: Again, just pass `(NULL, 0, st)`

adam900710: Although you will also need to remove the `UASSERT()` from `hardlink_compare_nodes()`.

mkfs/rootdir.c
daradib: Should we delete the `UASSERT(entry1->root)` and `UASSERT(entry2->root)` in `hardlink_compare_nodes`?

mkfs/rootdir.c
daradib: The existing assertion in `hardlink_compare_nodes` would prevent that from compiling.

adam900710: Yes, remove them. As that infrastructure was introduced when that `root` should always be a valid one. But it's no longer the case anymore.

daradib: Sounds good.

mkfs/rootdir.c
```diff
@@ -2213,17 +2211,36 @@ static void add_data_size(const char *fpath, const struct stat *st)
 static int ftw_add_entry_size(const char *fpath, const struct stat *st,
 			      int type, struct FTW *ftwbuf)
 {
+	int ret;
```
adam900710, line 2214: This `ret` is only utilized inside that `if (S_ISREG())` branch, I'll move the definition inside that branch during merge.

No need to refresh the branch.

adam900710 (approved):

PR #1144, opened 2026-07-13
btrfs-progs: check: allow checksummed extents for NODATASUM inodes

daandemeyer:
Checksummed extents can be shared by checksummed and NODATASUM inodes
after reflinking from the former into the latter. This is supported by
the kernel series below.

Link: https://lore.kernel.org/linux-btrfs/20260712-reflink-into-nodatasum-v1-0-9f27ef836073@amutable.com/
Signed-off-by: Daan De Meyer <daan@amutable.com>

### DIFF
--- a/check/main.c
+++ b/check/main.c
@@ -881,8 +881,10 @@ static void maybe_free_inode_rec(struct cache_tree *inode_cache,
 	}
 
 	if (S_ISREG(rec->imode) || S_ISLNK(rec->imode)) {
-		if (rec->found_csum_item && rec->nodatasum)
-			rec->errors |= I_ERR_ODD_CSUM_ITEM;
+		/*
+		 * A NODATASUM inode may reference a checksummed extent cloned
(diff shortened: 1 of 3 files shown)


adam900710:
Nope, all your changes are just masking any file extent with checksum for a nodatasum inode.

This will make btrfs to ignore all kinds of csum related bugs.

A big NACK.

daandemeyer:
I was gonna make this into a draft just to have the code available for review, I have to say the outright hostility is rather off putting for a new contributor.

adam900710:
I'd say screwing up csums bugs detection is a more hostile move for a tool that is designed to detect such bugs.

If you are not sure if it's a good idea, send everything as an RFC first.
If this compatibility problem never came up to you, then I hope you will at least remember this incident and prevent similar problems in the future.

And ignoring a bug report for almost a month without any response, is not going to give anyone a good impression.

PR #1147, opened 2026-07-28
Error handling fixes and cleanups

kdave:
- **btrfs-progs: handle strdup() failures**
- **btrfs-progs: don't reuse filesystem fd for sysfs files**
- **btrfs-progs: reset errno before strtoull() when parsing numbers**
- **btrfs-progs: use parse_u64() in cmd_subvolume_sync()**
- **btrfs-progs: simplify reading read partition size in device_get_partition_size_sysfs()**
- **btrfs-progs: use path API to build subvolume paths**
- **btrfs-progs: use full size path buffer in cmd_subvolume_delete()**
- **btrfs-progs: fix assigning negative errno**
- **btrfs-progs: receive: handle writes of 0 bytes**
- **btrfs-progs: receive: use correct type for pwrite in process_write()**
- **btrfs-progs: drop type casts from set/get_label calls**

### DIFF
--- a/cmds/subvolume-list.c
+++ b/cmds/subvolume-list.c
@@ -728,6 +728,8 @@ static int resolve_root(struct rb_root *rl, struct root_info *ri,
 			len += add_len + 1;
 		} else {
 			full_path = strdup(found->path);
+			if (!full_path)
+				return -ENOMEM;
 			len = add_len;
 		}
 		if (!ri->top_id)
--- a/cmds/subvolume.c
+++ b/cmds/subvolume.c
@@ -1808,9 +1808,8 @@ static int cmd_subvolume_sync(const struct cmd_struct *cmd, int argc, char **arg
 			const char *arg;
 
 			arg = argv[optind + i];
-			errno = 0;
-			id = strtoull(arg, NULL, 10);
-			if (errno) {
+			ret = parse_u64(arg, &id);
+			if (ret < 0) {
 				error("unrecognized subvolume id %s", arg);
 				ret = 1;
 				goto out;
--- a/cmds/subvolume.c
+++ b/cmds/subvolume.c
@@ -457,8 +457,18 @@ again:
 		goto out;
 	}
 	dupdname = strdup(cpath);
+	if (!dupdname) {
+		error_mem("duplicate path");
+		ret = -ENOMEM;
+		goto out;
+	}
 	dname = path_dirname(dupdname);
 	dupvname = strdup(cpath);
+	if (!dupvname) {
+		error_mem("duplicate path");
+		ret = -ENOMEM;
+		goto out;
+	}
 	vname = path_basename(dupvname);
 	free(cpath);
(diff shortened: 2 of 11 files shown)


common/device-utils.c
```diff
@@ -342,17 +340,11 @@ static int device_get_partition_size_sysfs(const char *dev, u64 *size_ret)
 	ret = path_cat3_out(sysfs, "/sys/class/block", name, "size");
 	if (ret < 0)
 		return ret;
-	sysfd = open(sysfs, O_RDONLY);
-	if (sysfd < 0)
-		return -errno;
-	ret = sysfs_read_file(sysfd, sizebuf, sizeof(sizebuf));
-	close(sysfd);
+
+	ret = sysfs_read_file_u64(sysfs, &size);
```
adam900710: `sysfs` is the full sysfs string, aka, `/sys/class/block/<dev>/size`, but `sysfs_read_file_u64()` itself will append `/sys/fs/btrfs` to the beginning of the path, through `sysfs_open_file_flags()` helper.

So `sysfs_read_file_u64()` using that path created by `path_cat3_out()` will always fail with `-ENOENT`.

Since we're reading from non-btrfs sysfs directory, we can not use `sysfs_read_file_u64()` here.

cmds/subvolume.c
```diff
@@ -729,28 +730,18 @@ static int cmd_subvolume_snapshot(const struct cmd_struct *cmd, int argc, char *
 		}
 		newname = path_basename(dupname);
 
-		dstdir = malloc(strlen(dst) + 1 + strlen(newname) + 1);
-		if (!dstdir) {
-			error_mem(NULL);
-			free(dupname);
+		ret = path_cat_out(dest_dir, dst, newname);
+		if (ret < 0) {
+			error("path too long");
```
adam900710, line 736: Missing error handling for `dupname`.

cmds/subvolume-list.c
```diff
@@ -728,6 +728,10 @@ static int resolve_root(struct rb_root *rl, struct root_info *ri,
 			len += add_len + 1;
 		} else {
 			full_path = strdup(found->path);
+			if (!full_path) {
+				free(full_path);
```
adam900710: This is just `free(NULL);` which is duplicated.

common/device-utils.c
kdave: Ah right, I'll drop the patch.

cmds/subvolume.c
kdave: free(dupname) moved after path_cat_out, after that it's not needed.

cmds/subvolume-list.c
kdave: Removed.

kdave:
Update pushed:
- removed btrfs-progs: simplify reading read partition size in device_get_partition_size_sysfs()
- free() calls fixed

PR #1154, opened 2026-08-17
btrfs-progs: docs: add extra LLM disclosure following kernel guidelines

adam900710:
There are increased LLM assisted bug fixes in the mailing list, and I have a very strong feeling that most of the new bug fixes from unknown developers are at least assisted of LLM, and they never properly disclose the usage of LLM.

I don't know much difference such updated docs will make, but at least let's try to make them disclose the LLM usage.

### DIFF
--- a/Documentation/btrfs-quota.rst
+++ b/Documentation/btrfs-quota.rst
@@ -22,12 +22,23 @@ of a btrfs filesystem. The quota groups (qgroups) are managed by the subcommand
 STABILITY AND PERFORMANCE IMPLICATIONS
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The qgroup implementation is considered reasonably stable for daily use and has
-been enabled in various distributions.
+The qgroup mode is considered not recommended for daily usage, unless there is
+no planned new snapshot/subvolume creation and deletion.
 
-When quotas are activated, they affect all extent processing, which takes a
-performance hit. Activation of qgroups is not recommended unless the user
-intends to actually use them.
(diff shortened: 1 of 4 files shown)


kdave:
At this moment I don't care much if AI/LLM is used for patches, they get the same level of review, new/unknown people or not. This is a very divisive topic, I don't have as strict requirements as kernel but at least for licensing and human behind the actual pull request or submission it makes sense.

Mentioning which tool/model was used has little value other than giving credit, or advertising. Unlike other targeted tools like coccinelle or sparse, the AI models can do anything, unless the patch generation is reproducible it does not help us much.

adam900710:
This doc is not only for btrfs-progs, but also kernel contributions.
And I'm not asking for anything extra, but only following the existing doctrine.

The reasons I want to proper LLM disclosure are:

- LLM generated patch has a very high chance of duplication
  And none of them are properly checking btrfs' for-next.
  So it's already not the first time some one sends out a fix that is already fixed better in for-next.

- Some authors even seems to use LLM to generate replies to review
  Which has all the problem of the LLM, it always tend to agree to whatever human forces,
  even if the review is incorrect.

  Such authors should not be trusted as they can not defend whatever they submitted.

Sure, we keep the same level of review no matter if it's generated by human or LLM, but the trend is already there, those new/unknown authors are completely relying on LLM to expose bugs with little ability to defend whatever they submit.
And they tend to intentionally hide the LLM disclosure, to act like if they find the bug by themselves and understand what they are doing.

That's why I want to proper LLM disclosure other than not.

I do not care about what model they are using, I only care about if they are some LLM-era script kiddies.

PR #1157, opened 2026-08-20
Set creation/change time (otime/ctime) for offline-created subvolumes and `convert`d root subvolume

koraynilay:
Fix `subvol->otime` and `subvol->ctime` not being set on:
- the root subvolume of a `btrfs-convert`d filesystem
- subvolumes created by `mkfs.btrfs` using `--rootdir` and `--subvol` options
- (less important) the `ext2_saved` subvolume of a `btrfs-convert`'d filesystem

Thanks

### DIFF
--- a/common/root-tree-utils.c
+++ b/common/root-tree-utils.c
@@ -15,6 +15,7 @@
  */
 
 #include "kerncompat.h"
+#include <string.h>
 #include <time.h>
 #include <uuid/uuid.h>
 #include "kernel-shared/disk-io.h"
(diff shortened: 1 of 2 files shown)


convert/common.c
```diff
@@ -20,6 +20,7 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
```
adam900710, line 23: No need to include time.h explicitly, it's already included by some other headers.

common/root-tree-utils.c
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "kerncompat.h"
+#include <string.h>
```
adam900710, line 18: No need to include that header explicitly.

There are already users of memset() in the same file, thus it's already included by some other headers.

adam900710 (approved):
Only very tiny issues related to headers.

Otherwise looks good to me.

You do not need to refresh the patches, I will change them during merge.

koraynilay:
Thanks, btw I added the headers because I checked other files and e.g. `cmds/subvolume.c` has an explicit `#include <time.h>` that was introduced in commit 40c4ba74ec09f60228d80c2e5ff77f824336256a because of IWYU (include what you use) so I thought it would be cleaner.

adam900710:
Normally I only add headers when the LSP server is showing errors or unable to complete the function.

It's already pretty messy to manage the headers for C, so I tend to not include extra unless it failed to compile.

Overall you don't need to bother as long as the project still compiles.

Issue #1158, opened 2026-08-21
libbtrfsutil: second call to btrfs_util_subvolume_iterator_next_info() after an error reads out of bounds


wehrwolfmann:
Calling btrfs_util_subvolume_iterator_next_info() again after it has already returned an error makes libbtrfsutil
read far outside its search buffer, so instead of the same error back I get a SIGSEGV. I found it through
btrfs-assistant, which loops until it sees STOP_ITERATION and so dies for an unprivileged user before it can print
its own --help; that project has a patch from me too. Looping on a failed iterator is questionable use of the API,
but the docs for btrfs_util_subvolume_iter_next() only promise a "non-zero error code on failure" and never say
the iterator has to be dropped afterwards, so I would expect the second call to fail, not to crash.

An unprivileged caller passing top = BTRFS_FS_TREE_OBJECTID still gets use_tree_search, since create picks tree
search whenever top != 0, so the iterator is created fine and only the first next() fails. In
subvolume_iterator_next_tree_search() the else branch sets top->search.key.nr_items = 4096 before
BTRFS_IOC_TREE_SEARCH and, when the ioctl fails, returns BTRFS_UTIL_ERROR_SEARCH_FAILED right there, leaving
nr_items at 4096 with items_pos and buf_off still 0. The next call finds items_pos < nr_items, takes the
consume-an-item branch and parses search headers out of a buffer that the ioctl never filled.

A probe on the iterator's internals shows items_pos at 4096 and buf_off at 131072 after that second call, 128 KB
past the start of a 3992-byte buffer. It survived the walk here only because the memory behind the buffer happened
to be mapped and zeroed; on a dirty heap the length fields are garbage and it faults at once. When it does
survive, items_pos ends up equal to nr_items, so every later call reissues the ioctl and returns 12 forever.

Restoring nr_items (or setting items_pos = nr_items) on that error path before returning would make the second
call fail exactly like the first. libbtrfsutil/subvolume.c in devel is byte for byte identical to v7.1, so nothing
has changed since. Attached are a reproducer that segfaults on the second call, the probe that prints the internal
state, and the output of both; btrfs-progs 7.1 on Arch/CachyOS, kernel 7.2.0, x86_64, filesystem mounted at /.

wehrwolfmann:
Attachment: iterator-loop-reproducer.c

```c
/*
 * libbtrfsutil: calling btrfs_util_subvolume_iterator_next_info() again after
[...]

PR #1159, opened 2026-08-22
libbtrfsutil: don't leave the subvolume iterator pointing at an unfilled search buffer

wehrwolfmann:
Fixes the out-of-bounds read reported in #1158.

`subvolume_iterator_next_tree_search()` sets `nr_items` to 4096 before `BTRFS_IOC_TREE_SEARCH` and returns `BTRFS_UTIL_ERROR_SEARCH_FAILED` without restoring it when the ioctl fails. `items_pos` and `buf_off` are left at 0, so the next call to `btrfs_util_subvolume_iterator_next()` takes the `items_pos < nr_items` branch and parses search headers out of a buffer the failed ioctl never filled. With 4096 phantom items the walk ends roughly 128 KB past a 3992 byte buffer, which either segfaults or spins.

An unprivileged caller reaches this without doing anything unusual: `use_tree_search` is true whenever `top != 0`, so the iterator is created successfully and the first ioctl then fails with EPERM. That is how btrfs-assistant crashed on `--help`.

The patch resets the entry on the error path, so the buffer is never parsed before an ioctl has filled it.

Checked on btrfs-progs devel, kernel 7.2.0, one single-device filesystem at /, using the reproducer attached to #1158 and a bounded variant of it:

| | packaged 7.1 | this branch |
|---|---|---|
| ordinary user | SIGSEGV (exit 139) | `create -> 0`, then `12 (Could not search B-tree)` on every call, exit 0 |
| root | `@`, `@/var/lib/portables`, `@/var/lib/machines`, `@/.snapshots`, `@/.snapshots/950/snapshot` | identical, byte for byte |

The success path is untouched, only the error path changes.

I also have a follow-up patch that remembers the failure in the iterator so a caller which loops until `STOP_ITERATION` stops re-issuing the ioctl that just failed. It does not save such a caller from looping, so I left it out here; happy to send it if you want it.

### DIFF
--- a/libbtrfsutil/subvolume.c
+++ b/libbtrfsutil/subvolume.c
@@ -1501,8 +1501,18 @@ static enum btrfs_util_error subvolume_iterator_next_tree_search(struct btrfs_ut
 			} else {
 				top->search.key.nr_items = 4096;
 				ret = ioctl(iter->fd, BTRFS_IOC_TREE_SEARCH, &top->search);
-				if (ret == -1)
+				if (ret == -1) {
+					/*
+					 * nr_items was set above but the ioctl
+					 * left buf untouched. Clear it so that
+					 * a caller which keeps iterating does
+					 * not walk the stale buffer.
+					 */
+					top->search.key.nr_items = 0;
+					top->items_pos = 0;
+					top->buf_off = 0;
 					return BTRFS_UTIL_ERROR_SEARCH_FAILED;
+				}
 				top->items_pos = 0;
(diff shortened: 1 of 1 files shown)


adam900710:
The fix looks good to me.

But please disclose the LLM usage through "Assisted-by: LLM" tag, as I believe both the reproducer and fix is generated by LLM.

wehrwolfmann:
Tag added and force-pushed, the branch is now at a2901cd.

Thanks for the review.

PR #1161, opened 2026-08-26
btrfs-progs: sb-mod: use per-type digest size for superblock csum

### LINKED ISSUE #1160: btrfs-sb-mod: checksum computed/written with CRC32 size for all csum types (corrupts xxhash64+ superblocks on write)
## Summary

`btrfs-sb-mod` verifies and rewrites the superblock checksum using a single global size initialized to the **CRC32** digest size, regardless of the filesystem's actual `csum_type`. On filesystems using xxhash64 (or sha256/blake2), every write produces a **truncated, invalid superblock checksum**, making the (repaired) filesystem unmountable. Read-only verification also misreports.

## Affected code

`btrfs-sb-mod.c` (confirmed identical on tag `v7.1` and current `devel`):

- Line 33: `static int csum_size;`
- Line 377: `csum_size = btrfs_csum_type_size(BTRFS_CSUM_TYPE_CRC32);` — set once, never per-type
- Line ~43 (`check_csum_superblock()`): `return (memcmp(sb, result, csum_size) == 0);`
- Line ~57 (`update_block_csum()`): `memcpy(&hdr->csum, result, csum_size);` after zeroing the 32-byte field

## Impact

On an xxhash64 filesystem, after any modifying fieldspec (`=`/`+`/`-`/`^`/`@`), `update_block_csum()` writes only 4 of the 8 xxhash64 bytes into a zeroed 32-byte csum field. `btrfs inspect-internal dump-super` then reports `[DON'T MATCH]` and the kernel rejects the superblock. Example observed on a real recovery case:

```
csum                 0x7cbc602b00000000 [DON'T MATCH]   # before fix
csum                 0x7cbc602bf714f3a8 [match]         # after fix below
```

## Reproduction

```sh
truncate -s 64M dev.img                       # must be large enough for offset 65536
mkfs.btrfs --csum xxhash64 -r fs.img          # any small fs image
dd if=fs.img of=dev.img bs=4096 seek=16 count=1 conv=notrunc   # place SB at BTRFS_SUPER_INFO_OFFSET
./btrfs-sb-mod dev.img generation '=1'
btrfs inspect-internal dump-super dev.img | grep csum   # -> [DON'T MATCH]
```

(With `--csum crc32c` the same commands report `[match]`, which is why the bug hides from default-config testing.)

## Suggested fix

Size both operations by the block's own csum type instead of the global:

```c
static int check_csum_superblock(void *sb)
{
	u16 csum_type = btrfs_super_csum_type(sb);
	int csize = btrfs_csum_type_size(csum_type);
	...
	return (memcmp(sb, result, csize) == 0);
}

static void update_block_csum(void *block)
{
	u16 csum_type = btrfs_super_csum_type(block);
	int csize = btrfs_csum_type_size(csum_type);
	...
	memcpy(&hdr->csum, result, csize);
}
```

I have this as a tested one-file patch against v7.1 (verified `[match]` via dump-super on patched scratch images for xxhash64) and can send it inline or as a pull request if preferred.

## Environment

- Arch Linux x86_64, btrfs-progs v7.1 sources; `devel` branch checked today shows the same code
- Found while packaging/building `btrfs-sb-mod` to repair a real corrupted (xxhash64) filesystem

HaleTom:
## Summary

`btrfs-sb-mod` hardcodes the superblock checksum length to the CRC32C
digest size instead of using the digest size of the filesystem's actual
checksum type:

- `main()` does `csum_size = btrfs_csum_type_size(BTRFS_CSUM_TYPE_CRC32);`
  while every read/write path derives `csum_type` from
  `btrfs_super_csum_type(sb)`.
- `check_csum_superblock()` then compares only the first
  `csum_size` (= 4) bytes.
- `update_block_csum()` zeroes the whole 32-byte csum field and writes
  back only the first 4 bytes of the digest.

## Affected algorithms

| csum type | digest size | affected |
|---|---|---|
| crc32c | 4 | no (only case where 4 is correct) |
| xxhash64 | 8 | yes — stored csum truncated, e.g. `0x7cbc602b00000000` |
| sha256 | 32 | yes |
| blake2b | 32 | yes |

So three of the four supported algorithms produce `[DON'T MATCH]`
superblocks after any `btrfs-sb-mod` write, and reads mis-validate.

## Fix

Replace both uses of the global with
`btrfs_csum_type_size(csum_type)` (same helper the rest of btrfs-progs
uses) and drop the now-unused global. This fixes #1160.

## Testing

Hit during a real recovery of a 5 TB btrbk target (xxhash64): with the
patched binary,

```
$ btrfs-sb-mod img generation '=26549' root '=761057165312'
...
Update csum
$ btrfs inspect-internal dump-super img | grep csum
csum            0x7cbc602bf714f3a8 [match]
```

vs unpatched: `csum 0x7cbc602b00000000 [DON'T MATCH]`. The repaired
image mounts read-only and the kernel accepts the superblock.

Fixes #1160

### DIFF
--- a/btrfs-sb-mod.c
+++ b/btrfs-sb-mod.c
@@ -30,7 +30,6 @@
 
 #define BLOCKSIZE (4096)
 static char buf[BLOCKSIZE];
-static int csum_size;
 
 static int check_csum_superblock(void *sb)
 {
@@ -40,7 +39,7 @@ static int check_csum_superblock(void *sb)
 	btrfs_csum_data(csum_type, (unsigned char *)sb + BTRFS_CSUM_SIZE,
 			result, BTRFS_SUPER_INFO_SIZE - BTRFS_CSUM_SIZE);
 
-	return (memcmp(sb, result, csum_size) == 0);
+	return (memcmp(sb, result, btrfs_csum_type_size(csum_type)) == 0);
 }
 
 static void update_block_csum(void *block)
@@ -54,7 +53,7 @@ static void update_block_csum(void *block)
 
 	memset(block, 0, BTRFS_CSUM_SIZE);
 	hdr = (struct btrfs_header *)block;
-	memcpy(&hdr->csum, result, csum_size);
+	memcpy(&hdr->csum, result, btrfs_csum_type_size(csum_type));
 }
 
 static u64 arg_strtou64(const char *str)
@@ -159,7 +158,7 @@ struct sb_field {
 			printf("SET: "#fname" "f_dec" (0x"f_hex")\n", \
 			(f_type)*val, (f_type)*val);				\
 			sb->fname = cpu_to_le##bits(*val);			\
-		} else {							\
+		} else {						\
 			*val = le##bits##_to_cpu(sb->fname);			\
 			printf("GET: "#fname" "f_dec" (0x"f_hex")\n", 	\
 			(f_type)*val, (f_type)*val);			\
@@ -374,7 +373,6 @@ int main(int argc, char **argv)
 	}
 
 	/* verify superblock */
-	csum_size = btrfs_csum_type_size(BTRFS_CSUM_TYPE_CRC32);
 	off = BTRFS_SUPER_INFO_OFFSET;
 
 	ret = pread(fd, buf, BLOCKSIZE, off);
--- a/btrfs-sb-mod.c
+++ b/btrfs-sb-mod.c
@@ -158,7 +158,7 @@ struct sb_field {
 			printf("SET: "#fname" "f_dec" (0x"f_hex")\n", \
 			(f_type)*val, (f_type)*val);				\
 			sb->fname = cpu_to_le##bits(*val);			\
-		} else {						\
+		} else {							\
 			*val = le##bits##_to_cpu(sb->fname);			\
 			printf("GET: "#fname" "f_dec" (0x"f_hex")\n", 	\
 			(f_type)*val, (f_type)*val);			\


adam900710:
I believe you should have already noticed this PR doesn't pass CI.
And that's caused by the fuzz tests where `btrfs-sb-mod` is called to modify csum_type to invalid value (-1).

This means, with your PR, `btrfs-sb-mod` is no longer able to create create such corrupted image.

I do not think we really need that ability, but it's still a regression.

I think you should add special handling to support invalid csum types (e.g. warning, only update csum_type, but do not recalculate/touch the csum).

btrfs-sb-mod.c
```diff
@@ -158,7 +158,7 @@ struct sb_field {
 			printf("SET: "#fname" "f_dec" (0x"f_hex")\n", \
 			(f_type)*val, (f_type)*val);				\
 			sb->fname = cpu_to_le##bits(*val);			\
-		} else {						\
+		} else {							\
```
adam900710, line 161: Firstly, missing `btrfs-progs:` prefix to the subject, secondly we do not accept simple typo/whitespace fixes.

HaleTom: I thought you'd squash all the commits in this PR into one?

This is a fix of https://github.com/kdave/btrfs-progs/pull/1161/changes/727a29f0d738ebde9d41efafae895ae0b44a8e25

Would you prefer I force-push this PR as one commit?

kdave:
I've added the checksum validation and removed the `csum_type` from fuzzing. If an invalid value is set from the parameter the reasonable thing is to exit with error. Besides the minor whitespace fixes, the other patch makes sens and I've added it to devel.

Issue #1163, opened 2026-08-30
btrfs check: invalid key types are printed but not counted — "no error found", exit 0, on a filesystem the kernel forces read-only


atahan150:
## Summary

Two findings. I am leading with the second one because it is the one that can be
verified from the source alone, and because it is what turned a repairable problem
into a second failure a day later.

1. **`btrfs check` detects invalid key types, prints them, and then discards the
   error.** The run still ends with `no error found` and **exit status 0**, even
   though the kernel refuses to write to the filesystem. This is unconditional and
   is visible in `check/main.c` today.

2. On this particular filesystem the four offending keys appeared **during
   `btrfs check --repair`**. Four `(TREE_LOG_OBJECTID, ROOT_ITEM, <subvolid>)` items
   ended up at the end of an `EXTENT_TREE` leaf. I have a strong before/after
   correlation for this but no `btrfs-image` of the pre-repair state, so I present it
   as an observation rather than a conclusion — see the caveats at the end.

## Environment

- btrfs-progs **v7.1** (current latest release; the relevant code is byte-identical
  in `devel` as of 2026-08-30)
- Kernels: `7.2.2-1-cachyos` (host), `6.18.42-1-cachyos-lts` (rescue/live, where the
  read-only abort was captured)
- Single device btrfs inside LUKS2, 830.93 GiB
- `nodesize 16384`, `sectorsize 4096`, `csum_type 0 (crc32c)`, `incompat_flags 0x371`
  (`MIXED_BACKREF | COMPRESS_ZSTD | BIG_METADATA | EXTENDED_IREF | SKINNY_METADATA | NO_HOLES`)
- data `single`, metadata `DUP`
- 10 subvolumes + 51 snapper snapshots

---

# Finding 1 — invalid key types are reported but not counted as errors

`check_type_with_root()` correctly rejects a `ROOT_ITEM` key that is not in the root
tree (`check/main.c`, v7.1 line 6431, unchanged in `devel`):

```c
	case BTRFS_ROOT_ITEM_KEY:
		if (rootid != BTRFS_ROOT_TREE_OBJECTID)
			goto err;
		break;
...
err:
	report_mismatch_key_root(key_type, rootid);
	return -EINVAL;
```

The caller in `run_next_block()` throws that `-EINVAL` away
(`check/main.c`, v7.1 lines 6604-6608, unchanged in `devel`):

```c
			if (check_type_with_root(btrfs_header_owner(buf),
						 key.type)) {
[...]

Issue #1014, opened 2025-08-09
BTRFS error (device sdc): tree first key mismatch detected


kocoten1992:
Hi, I'm converting my btrfs from raid6 to raid10, but currently I'm encouter this error:

```
[ 4386.497028] BTRFS error (device sdc): tree first key mismatch detected, bytenr=32740860231680 parent_transid=908625 key expected=(18446744073709551606,128,33687613767680) has=(36324653895680,168,184320)
[ 4386.497249] BTRFS error (device sdc): tree first key mismatch detected, bytenr=32740860231680 parent_transid=908625 key expected=(18446744073709551606,128,33687613767680) has=(36324653895680,168,184320)
[ 4386.497286] BTRFS: error (device sdc) in btrfs_replay_log:2104: errno=-5 IO failure (Failed to recover log tree)
[ 4386.531770] BTRFS error (device sdc state E): open_ctree failed: -5
```

Base on my google search, the suggestion was to backup and move to data to other filesystem and reinstall btrfs.

But is there a way to save this filesystem? I've not run `btrfs check --repair` yet.

Thank you very much!

Zygo:
This is a log tree failure.  Assuming nothing else is wrong, run `btrfs rescue zero-log` and it should be fine.  You'll lose the last 30 seconds of `fsync`ed data (assuming default `commit=` option).

kocoten1992:
Hi, I've run btrfs check, this is the result:

<details><summary>Details</summary>
<p>

root@d4:/mnt# btrfs check /dev/sdc
Opening filesystem to check...
Checking filesystem on /dev/sdc
UUID: b802cbcc-f8b2-4d7a-b741-5dc66c3790ad
[1/8] checking log
[2/8] checking root items
[3/8] checking extents
Invalid key type(EXTENT_ITEM) found in root(CSUM_TREE)
ignoring invalid key
Invalid key type(EXTENT_ITEM) found in root(CSUM_TREE)
ignoring invalid key
Invalid key type(EXTENT_ITEM) found in root(CSUM_TREE)
ignoring invalid key
Invalid key type(EXTENT_ITEM) found in root(CSUM_TREE)
ignoring invalid key
Invalid key type(EXTENT_ITEM) found in root(CSUM_TREE)
ignoring invalid key
Invalid key type(EXTENT_ITEM) found in root(CSUM_TREE)
ignoring invalid key
Invalid key type(EXTENT_ITEM) found in root(CSUM_TREE)
[...]

PR #1175, opened 2026-09-19
btrfs-progs: check: report invalid key types as an error

swimmingbrain:
In the original mode, `run_next_block()` checks every key type against the owner of its leaf. When the type does not belong there, it prints the key, skips the item and carries on, and nothing remembers that it happened. So the run ends with `no error found` and exit status 0. In #1163 that happened on a filesystem the kernel had already forced read-only because of those very items, and #1014 shows the same two lines passing as harmless noise.

```
Invalid key type(DEV_EXTENT) found in root(EXTENT_TREE)
ignoring invalid key
found 163840 bytes used, no error found
```

Skipping the item is still right for the walk, so that stays. A flag now records it, the same way `found_unknown_key` does for unknown inode keys, and `check_chunks_and_extents()` returns an error at the end of the pass when it is set. The flag is cleared at `again:` so a repair loop reports the last pass and not an earlier one.

The test image is a fresh filesystem with one DEV_EXTENT item appended to the extent tree leaf, well formed so the tree checker lets it through, checksum fixed up. On devel it passes check with exit 0. With the change:

```
Invalid key type(DEV_EXTENT) found in root(EXTENT_TREE)
ignoring invalid key
ERROR: errors found in extent allocation tree or chunk allocation
found 163840 bytes used, error(s) found
```

The new case is `fsck-tests/076-invalid-key-type`. It fails on devel with "unexpected success" and passes with the change, in both modes. Of the other fsck tests, the 62 that work on images alone pass here; the remaining 11 need a btrfs mount or loop devices, which my container does not have.

Lowmem mode already fails on this image, for a different reason, it cannot find a chunk for the dev extent. It has no key type check of its own, I left that alone.

Not touched, but I noticed it on the way: `check_type_with_root()` has `case BTRFS_CSUM_TREE_OBJECTID:` in a switch over key types, which looks like it was meant to be `BTRFS_EXTENT_CSUM_KEY`.

Issue: #1163


check/main.c
```diff
@@ -90,6 +90,7 @@ bool check_data_csum = false;
 static bool check_qgroup_accounting = true;
 static bool found_free_ino_cache = false;
 static bool found_unknown_key = false;
+static bool found_invalid_key_type = false;
```
adam900710: You are only doing the check in the original mode, why not on lowmem mode?

You mentioned the test case is already rejected by lowmem mode, but that's not a valid reason.
It only worked for that specific case, not always the case for every case.

tests/fsck-tests/076-invalid-key-type/test.sh
```diff
@@ -0,0 +1,17 @@
...
+#
+# Verify that check fails on a key whose type does not belong in the tree it
+# was found in. It used to print the key, skip it and report no error.
+
+source "$TEST_TOP/common" || exit
+
+check_prereq btrfs
+
+check_image() {
+	# The image has a DEV_EXTENT item at the end of the extent tree leaf,
```
adam900710: This is not a good one, as you have already mentioned lowmem rejects it by a different reason.

Either enhance the DEV_EXTENT item so that it can pass unpatched lowmem mode, or use some other valid items.

